PR #51, opened 2024-03-01
Adds a starter for a snapcraft.yaml

mz2:
Adds a snap build configuration with at least some of the right interface requirements that will be needed.

Based on (very) cursory testing of the following kind, Works On My Machine:

```bash
❯ hwctl
NotSeen(
    NotSeenResponse {
        status: "Not Seen",
    },
)
```

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -4,7 +4,7 @@ version: '0.1' # rewritten with the help of `adopt-info` below
 adopt-info: hwctl
 summary: Hardware API client
 description: Canonical Certified Hardware API client
-  
+
 grade: stable
 confinement: strict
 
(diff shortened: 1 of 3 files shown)


snap/snapcraft.yaml
```diff
@@ -0,0 +1,35 @@
+name: hwctl
+base: core22
+version: '0.1' # rewritten with the help of `adopt-info` below
```
nadzyah, line 3: right now the version (at least for the deb package and the lib) is `0.0.1`. should we keep it the same here?

mz2: The version to all artifacts should be coming from a git tag. I've fixed this up now for the snap build on this branch, can advise how to do that with the deb build separately.

nadzyah (approved):

PR #84, opened 2024-03-29
Implement basic logic for "Certified" status

nadzyah:
This PR closes [C3-623](https://warthogs.atlassian.net/browse/C3-629)

## Implementation

Here I've implemented the logic to determine a system we have seen, tested and certified. It works as "We have seen and certified the machine of the same vendor and its model contains an existing platform name (data in platform name parenthesis is ignored)". Here is an example of the response:

```
$ curl http://0.0.0.0:8080/v1/certification/status -X POST -H "Content-Type: application/json" \
          -d '{"vendor": "Dell", "model": "ChengMing 3980 (i3-9100)"}' -s | python3 -m json.tool
{
    "status": "Certified",
    "os": {
        "distributor": "Canonical Ltd.",
        "description": "",
        "version": "20.04",
        "codename": "focal",
        "kernel": {
            "name": "Linux",
            "version": "5.4.0-42-generic",
            "signature": "0000000"
        },
        "loaded_modules": []
    },
    "bios": {
        "firmware_revision": "1.0.0",
        "release_date": "2023-04-03",
        "revision": "A01",
        "vendor": "Lenovo",
        "version": "1.0.2"
    }
}
```

The logic has been implemented this way since, currently, we still certify machines in C3. It will probably need to be updated after we move to certifying configurations and after we implement the module for importing data from C3 to hwapi.

## Additional changes

The PR also introduces the data models and the script for populating the DB with sample data. It was created since for now we don't have any logic to import the data from C3, but already need to work with some data.

Additionally, now it's possible to stand up the development environment and run the tests using docker-compose. So now, instead of running all the checks by copying them from CI job, we can just execute the following command and run all the checks:

```bash
$ docker-compose up --attach-dependencies --force-recreate --abort-on-container-exit --build hwapi-test
```

The CI job was updated to use the `hwapi-test` container as well

## Testing

I've added a basic test that checks that if we have seen the machine of such vendor and model, we return the "Certified" response. If not, we should return "Not seen" response.

[C3-623]: https://warthogs.atlassian.net/browse/C3-623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/docker-compose.yml
```diff
@@ -0,0 +1,48 @@
...
+      bash -c '
+        python scripts/seed_db.py
+        black --check hwapi/ scripts/ tests/
+        ruff check hwapi/ scripts/ tests/
+        mypy hwapi/ scripts/ tests/
+        pytest'
+    working_dir: /home/app
+    volumes:
+      - .:/home/app
+      - ./data:/home/app/data
```
plars, line 31: Since this is for the hwapi-test, then you probably don't want to reuse data that could be there from a dev environment or even a previous test environment. And you also don't need to leave the data around. What do you think about creating a tmpfs volume so that you get a clean copy of the test data each time, and there's no chance it could interfere?
You could do that with:
```
    tmpfs:
      - /home/app/data
```

server/hwapi/data_models/models.py
```diff
@@ -0,0 +1,110 @@
...
+class Vendor(Base):
+    __tablename__ = "vendors"
+    name: Mapped[str] = mapped_column(unique=True, nullable=False)
+
+
+class Platform(Base):
+    __tablename__ = "platforms"
+    name: Mapped[str] = mapped_column(unique=True, nullable=False)
+    vendor_id: Mapped[int] = mapped_column(ForeignKey("vendors.id"))
+    vendor: Mapped[Vendor] = relationship(foreign_keys=[vendor_id])
```
omar-selo, line 43: Not sure if this form of relationship configuration works. Perhaps it would be safer to follow the SQLAlchemy [docs](https://docs.sqlalchemy.org/en/20/orm/basic_relationships.html)

server/hwapi/data_models/setup.py
```diff
@@ -0,0 +1,42 @@
...
+from .models import Base
+
+DB_URL = os.getenv("DB_URL", "sqlite://")
+engine = create_engine(DB_URL)
+SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
+
+DB_PATH = DB_URL.replace("sqlite:///", "")
+if DB_PATH and not os.path.exists(DB_PATH):
+    # Create tables with no data if DB path doesn't exist
+    Base.metadata.create_all(engine)
```
omar-selo, line 34: I guess this works right now in place of db migrations. But definitely wouldn't work if any schema was changed. So perhaps I'd replace it with migrations

server/hwapi/endpoints/certification/logic.py
```diff
@@ -0,0 +1,123 @@
...
+from hwapi.endpoints.certification.rbody_validators import (
+    CertificationStatusRequest,
+    CertifiedResponse,
+)
+from hwapi.data_models import models
+from hwapi.data_models import data_validators
+from hwapi.data_models.enums import CertificationStatus
+
+
+def is_certified(system_info: CertificationStatusRequest, db: Session):
```
omar-selo, line 34: would be nice if you could separate this into smaller functions that do one thing

server/hwapi/endpoints/certification/logic.py
```diff
@@ -0,0 +1,123 @@
...
+        .first()
+    )
+    if not platform:
+        return False, None
+
+    configurations = (
+        db.query(models.Configuration)
+        .filter(
+            # Check that system model contains configuration name
+            func.instr(system_info.model, models.Configuration.name) > 0,
```
omar-selo, line 58: I'm worried about the approach of checking names since we can really name configs and platforms something a bit different than what the model name is. Then this algorithm would assume that it's not certified. I had thought that we want to compare something more concrete and fixed

server/hwapi/endpoints/certification/logic.py
```diff
@@ -0,0 +1,123 @@
...
+
+    platform = (
+        db.query(models.Platform)
+        .filter(
+            # Check that system model contains platform name
+            func.instr(system_info.model, models.Platform.name) > 0,
+            models.Platform.vendor_id == vendor.id,
+        )
+        .first()
+    )
```
omar-selo, line 50: since this is an orm you can just access `vendor.platforms`. The same follows for the rest of the code. It seems a bit simpler and easier to read

server/scripts/seed_db.py
omar-selo, line 1: Can you please separate the logic here into small functions, makes it much easier to read :pray:

server/tests/endpoints/certification/test_certification.py
```diff
@@ -0,0 +1,69 @@
...
+
+def test_certified_status(test_client: TestClient):
+    response = test_client.post(
+        "/v1/certification/status",
+        json={"vendor": "Dell", "model": "ChengMing 3980 (i3-9100)"},
+    )
+
+    assert response.status_code == 200
+    expected_response = {
+        "status": "Certified",
```
omar-selo, line 33: how does this pass? I don't see any test data being inserted into the db. Perhaps it uses the seed data script? if so that would be coupling test cases to seed data. It would be better if the test cases were standalone and populate whatever data they need. Seed data should only run in dev database not test one. For tests we probably need to create a new test db for all of the tests, run db migrations on it and then for each test case we need to make sure to empty it out during the cleanup. Setting this up can be hairy so I'd recommend you look at how TO is doing it

omar-selo (changes requested):
Thanks for the work, overall I can see that this PR is doing quite a lot. Had a few comments, some of which might take a bit of time (testing setup and database migrations)

server/hwapi/data_models/models.py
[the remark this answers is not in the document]
nadzyah: I'm not sure whether we need migrations since the DB is planned to be a partial replica of C3 DB, so we'll just copy some data from C3 to our plan, and I believe will rewrite it periodically to keep the data up-to-date.
Anyway, I believe this decision will be made on the stage of implementing the logic of importing data from C3 to hwapi

.github/workflows/test_server.yaml
```diff
@@ -19,17 +19,34 @@ jobs:
       run:
         working-directory: server
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: "3.10"
-      - uses: Gr1N/setup-poetry@12c727a3dcf8c1a548d8d041c9d5ef5cebb3ba2e
+      - name: Checkout repo & submodules
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
nadzyah: I've removed this logic since we don't have any submodules. Should I add it here?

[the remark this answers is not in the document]
omar-selo: No need for it if we don't have submodules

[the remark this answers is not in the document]
nadzyah: I've added it just in case

server/hwapi/data_models/models.py
nadzyah: I've used the information from [this doc](https://docs.sqlalchemy.org/en/20/orm/join_conditions.html), but since our relations here are more obvious, we can use your suggestion here

server/hwapi/endpoints/certification/logic.py
nadzyah: For now, I'll remove the validation for configuration name, since @andrejvelichkovski is working on refactoring their naming in C3 DB. We'll adjust it after it's done

server/hwapi/endpoints/certification/logic.py
nadzyah: done

server/hwapi/data_models/models.py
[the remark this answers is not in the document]
omar-selo: Sounds good

server/hwapi/data_models/setup.py
omar-selo: I guess this is no longer relevant given your [answer above](https://github.com/canonical/hardware-api/pull/84#discussion_r1549391031)

omar-selo (changes requested):
Looks good, just having db indexes would be nice

server/hwapi/data_models/models.py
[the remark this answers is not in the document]
nadzyah: Which indices do you want to add? I believe ID field already has an index as PK, doesn't it?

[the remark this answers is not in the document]
nadzyah: since the DB will be used in read-only mode, do we actually need cascade deletions?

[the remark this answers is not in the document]
omar-selo: Nope shouldn't be needed then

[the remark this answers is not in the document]
nadzyah: I've added indices for FKs

PR #105, opened 2024-04-20
Publish the client app on Snap Store

mz2:
GitHub workflow for publishing the client app on Snap Store #105


nadzyah:
Closing due to #309 being merged

PR #111, opened 2024-04-26
Fix: remove top key from the hwlib response body

nadzyah:
This PR just removed the top-key like "Certified" or "Not seen" from the client response and outputs the value itself
So, instead of having `{'NotSeen':{'status':'Not Seen'}}`, we just get `{'status':'Not Seen'}`

The tests for both client and python bindings were updated as well

### DIFF
--- a/client/hwlib/tests/test_cert_status.rs
+++ b/client/hwlib/tests/test_cert_status.rs
@@ -1,5 +1,4 @@
 use hwlib::get_certification_status;
-use hwlib::models::rbody::CertificationStatusResponse;
 
 const SERVER_URL: &str = "https://example.com";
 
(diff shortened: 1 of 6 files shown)


client/hwlib/tests/test_cert_status.rs
```diff
@@ -1,34 +1,30 @@
 use hwlib::get_certification_status;
-use hwlib::models::rbody::CertificationStatusResponse;
 
 const SERVER_URL: &str = "https://example.com";
 
 #[tokio::test]
 async fn test_get_certification_status_not_seen() {
```
mz2, line 6: We are missing a test that checks against the serialised OpenAPI schema available in the repo.

mz2 (approved):
Fine to merge like this, but in follow-up we need a test that checks against the serialised OpenAPI schema, since it shouldn't be the API consuming team who finds departures from the schema. The problem is likely recurrent since we are going to want to continue to also test with this fake data in the future even with the real system available.

client/hwlib/tests/test_cert_status.rs
nadzyah: should this test be run for the client changes as well?

PR #113, opened 2024-04-26
Fix update_package_version script

nadzyah:
This PR just fixes the string format that we add to the debian/changelog file and fixes typos in readme

It also sets the package version to match the one on the LP (0.1.0 instead of 0.0.1)


Hook25 (approved):
+1, lgtm

PR #122, opened 2024-05-02
Updates for the charm

plars:
Once the charm built, I found a few things that would prevent it from installing and working properly. Working on fixes here. Still a WIP for the moment, but I'll take it out of draft once it's ready.


server/charm/charmcraft.yaml
```diff
@@ -15,3 +15,11 @@ config:
         Acceptable values are: "info", "debug", "warning", "error" and "critical"
       default: "info"
       type: string
+    port:
+      default: 30000
+      description: The port to listen on for incoming HTTP requests.
+      type: int
+    hostname:
+      default: hw
```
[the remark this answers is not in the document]
plars: This was just a default for dev instances if you don't set it. In reality, it's going to wind up being specified as hw.ubuntu.com for production and hw-staging.ubuntu.com for staging

server/charm/src/charm.py
```diff
@@ -3,62 +3,80 @@
...
-            self.unit.status = ops.BlockedStatus("invalid log level: '{log_level}'")
+            self.unit.status = ops.BlockedStatus(
+                "invalid log level: '{log_level}'"
+            )
+
+    @property
+    def _app_environment(self):
+        """Environment variables needed by the application."""
+        env = {}
+        return env
```
omar-selo, line 71: are we using this anywhere? or is it that we will be using it soon

server/terraform/hardware-api.tf
```diff
@@ -33,21 +39,37 @@ resource "juju_application" "hardware-api" {
   }
 
   config = {
-    hostname = var.environment == "staging" ? "hardware-api-staging.${var.external_ingress_hostname}" : "hardware-api.${var.external_ingress_hostname}"
+    hostname = var.external_ingress_hostname
```
omar-selo, line 42: we won't have another hostname for staging?

omar-selo (commented):
Looks good, just had a question on staging hostname

server/charm/src/charm.py
plars: It was already trying to use _app_environment but the property for it didn't exist yet. I thought about ripping it out, but I suspect we might want things like sentry at some point, so I thought it might be better to just add a stub for it for now.

omar-selo (approved):
Thanks for responding to my comments

PR #124, opened 2024-05-03
Terraform deploy updates

plars:
We can't yet use juju-terraform 0.12.0 because the controller is only at juju 2.9.46 at the moment.

0.10.1 was the latest one I could find that might work, but I ran into a lot of problems when it tried to deal with the model. Because we need our TLS secret imported ahead of time, we need to manually create the model. This is something we've easily dealt with before by using `terraform import ...` to set the terraform state for the model to point to the one we already created. However it couldn't work out the cloud/credentials data for some reason and wanted to delete the model that I already created. We can't let it do that!

Because we need to create the model manually anyway, there's no need to have it created as a juju_resource in the terraform plan. The only downside I can think of at the moment is that if you want to set this up in a dev environment where you might not care about TLS, then you'll need to manually create the model there also.


omar-selo (approved):
It's a shame that we couldn't keep the model in the terraform plan. Makes me even wonder if we should use an even older version of juju terraform provider. But this looks good to me

plars:
@omar-selo yeah, there might be some other approaches to this. At some point I'd like to dig deeper into understanding terraform and some of the other capabilities. Perhaps there's some other way to accomplish this that we haven't found yet. Once we have the controllers updated to at least 2.9.49 then we should be able to get to 0.12 at least

PR #131, opened 2024-05-09
Implement the certification status logic to match the spec

nadzyah:
This PR resolves [C3-676](https://warthogs.atlassian.net/browse/C3-676)

It also re-implements the logic for certified status and adds other statuses to match the following schema from [the spec](https://docs.google.com/document/d/1DY0xo9NJ5NBFf2R12G4qDUP_GCpkdmrSHlJgJn-s4F0/edit?pli=1)

```mermaid
graph TD
    A[Start: Check OEM] -->|No| B[Fail: report no certified images for the device]
    A -->|Yes| C[Match against board, chassis, BIOS information]
    C -->|No| D[Fail: report no certified images for the device]
    C -->|Yes| E["Are there disqualifying PCI / USB IDs or CPU codename (generation) mismatch?"]
    E -->|Yes| F[Related certified system exists:\n report non-certified status, list related certified systems & images]
    E -->|No| G[Check OS release / image metadata]
    G -->|No| H[Certified image exists:\n report non-certified status, list certified images for system]
    G -->|Yes| I[List certified images]
```

Since we don't have any option in C3 for now to disqualify devices for configs or machines, the 3rd step checks only CPU codename. Once this feature is implemented in C3, we'll update the logic to do such a check as well.

The tests were added to verify the logic for each step on the diagram

[C3-676]: https://warthogs.atlassian.net/browse/C3-676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/hwapi/endpoints/certification/certification.py
```diff
@@ -50,7 +62,14 @@ def check_certification(
...
     or some of its components have been seen on other systems)
     """
+    data: CertifiedResponse | RelatedCertifiedSystemExistsResponse | None
     certified, data = is_certified(system_info, db)
     if certified:
         return data
-    return NotCertifiedResponse(status=CertificationStatus.NOT_SEEN)
+    try:
+        partially_certifed, data = is_partially_certified(system_info, db)
+    except ValueError as exc:
```
plars, line 71: Do both is_certified() and is_partially_certified() need to be under this try/except block?

plars (commented):
This looks  good and lots of tests! I think it would be a gift to our future selves to be a little more descriptive in the comments for is_certified() and is_partially_certified() to make what we're looking for in each case really clear, especially since it might take some time to really trace down all of what is_partially_certified() is doing.

Otherwise, one small question below

server/hwapi/endpoints/certification/certification.py
nadzyah: For now, is_certified doesn’t raise any ValueErrors, but is_partially_certified does; that’s why only this method is wrapped with a try-except block. Do you think we should change it?

plars: Only if you think it might be able to raise that in the future. It just wasn't immediately clear to me if it was something we should be concerned with

server/hwapi/endpoints/certification/logic.py
```diff
@@ -46,27 +51,203 @@ def is_certified(system_info: CertificationStatusRequest, db: Session):
...
+    Check if some components of the system were seen on other certified machines.
+    """
+    release = (
+        repository.get_release_from_os(db, system_info.os) if system_info.os else None
+    )
+    board = repository.get_board_by_validator_data(db, system_info.board)
+    machines = repository.get_machines_with_same_hardware_params(
+        db, system_info.architecture, board, release
+    )
+    machine_ids = [machine.id for machine in machines]
```
omar-selo, line 91: small point here, but unless we specifically need a list then making a generator would be more memory efficient

server/hwapi/endpoints/certification/logic.py
```diff
@@ -46,27 +51,203 @@ def is_certified(system_info: CertificationStatusRequest, db: Session):
...
+        wireless=wireless_devices,
+        pci_peripherals=pci_devices,
+        usb_peripherals=usb_devices,
+    )
+
+
+def _get_matching_devices(
+    db: Session,
+    devices_query: Query,
+    device_validator_list: list,
```
omar-selo, line 231: why is `device_validator_list` a list

server/hwapi/endpoints/certification/logic.py
```diff
@@ -46,27 +51,203 @@ def is_certified(system_info: CertificationStatusRequest, db: Session):
...
+    :param validator_mapping: Mapping function to transform devices to validator objects.
+    :return: A list of matching device validator objects.
+    """
+    devices = []
+    for device_info in device_validator_list:
+        device = find_matching_function(db, devices_query, device_info)
+        if device is None:
+            continue
+        devices.append(validator_mapping(device, device_info))
+    return devices
```
omar-selo, line 253: This function is too complex. Personally, I don't know exactly what it does. Perhaps it's doing too many things (as I can tell by the number of parameters it takes). I'd would've tried to find a way to simplify it, but I didn't really understand it

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+        .first()
+    )
+    return board
+
+
+def get_devices_by_machine_ids(db: Session, machine_ids: list[int]) -> Query:
+    """
+    Return query that contains Devices joined with Reports and Certificates for a
+    given list of machine IDs
+    """
```
omar-selo, line 204: given that this function returns a query object. I'd rename it to something like `get_machines_devices_query`

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+    Find a CPU device across the given list of devices that contains
+    model name.
+    """
+    return devices_query.filter(
+        and_(
+            models.Device.name.like(f"%{cpu_validator.version}%"),
+            models.Device.category == DeviceCategory.PROCESSOR,
+        ),
+    ).first()
```
omar-selo, line 234: not too familiar with the device object or cpu data. But why is it that device name represents cpu version? Perhaps one needs to be renamed to the other.

Also couldn't like query produce false positives?

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+def find_matching_gpu(
+    db: Session, devices_query: Query, gpu_validator: device_validators.GPUValidator
+) -> models.Device | None:
+    """
+    Find a GPU device across the given list of devices that have the same model name and
+    identifier
+    """
+    return devices_query.filter(
+        and_(
+            models.Device.name == gpu_validator.version,
```
omar-selo, line 246: I'm seeing a pattern not sure why version is name

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+            models.Device.identifier == gpu_validator.identifier.lower(),
+            models.Device.category.in_(
+                [DeviceCategory.VIDEO.value, DeviceCategory.OTHER]
+            ),
```
omar-selo, line 250: I see some places where you check that device category is either what's expected or OTHER and other places where you don't check for OTHER. Is there any reason for that? if there is then we need to be more explicit with the reason.

Also it seems to me that OTHER category should be named UNKNOWN instead as it's not really OTHER if it's used this way

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+    network_validator: device_validators.NetworkAdapterValidator,
+) -> models.Device | None:
+    """
+    Find a network wired device across the given list of devices that has the same
+    idenitifier, model name, and bus
+    """
+    return devices_query.filter(
+        and_(
+            models.Device.identifier == network_validator.identifier.lower(),
+            models.Device.name == network_validator.model,
```
omar-selo, line 267: I see name is model for network devices :/

server/hwapi/data_models/repository.py
omar-selo, line 1: general comment about sqlalchemy, [Query is legacy/deprecated](https://docs.sqlalchemy.org/en/20/orm/queryguide/query.html#legacy-query-api) so ideally we shouldn't be introducing it in new code

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+    pci_validator: device_validators.PCIPeripheralValidator,
+) -> models.Device | None:
+    """
+    Find an arbitary pci device across the given list of devices that has the same
+    PCI ID and model name
+    """
+    return devices_query.filter(
+        and_(
+            models.Device.identifier == pci_validator.pci_id.lower(),
+            models.Device.bus == BusType.pci,
```
omar-selo, line 342: should `Device.bus` be renamed `Device.bus_type`?

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+    devices_query: Query,
+    usb_validator: device_validators.USBPeripheralValidator,
+) -> models.Device | None:
+    """
+    Find an arbitary usb device across the given list of devices that has the same
+    USB ID and model name
+    """
+    return devices_query.filter(
+        and_(
+            models.Device.identifier == usb_validator.usb_id.lower(),
```
omar-selo, line 359: This is a generic question. What is `identifier`?

server/hwapi/endpoints/certification/logic.py
```diff
@@ -46,27 +51,203 @@ def is_certified(system_info: CertificationStatusRequest, db: Session):
...
+            )
+            if board
+            else None
+        ),
+        processor=cpus,
+        gpu=gpus,
+        audio=audio_devices,
+        video=capture_devices,
+        network=network_devices,
+        wireless=wireless_devices,
```
omar-selo, line 222: The mapping between singular and plural is a bit weird. Would be great if we stick to one

Also it would be nice to have consistent vocabulary for instance either stick with `processor` or `cpu` and `video` or `capture_device`

omar-selo (changes requested):
Thanks for the PR, I can tell this was a difficult task. Also sorry for the delay, took much more time than I'd like to admit to review. I do have a lot of comments but they are mostly small

server/docker-compose.yml
```diff
@@ -23,17 +22,18 @@ services:
...
     command: >
       bash -c '
         black --check hwapi/ scripts/ tests/ &&
         ruff check hwapi/ scripts/ tests/ --no-cache &&
         mypy hwapi/ scripts/ tests/ &&
-        coverage run -m pytest tests/ -vv &&
-        coverage xml -o coverage/coverage.xml'
-    working_dir: /home/code
+        coverage run --source "hwapi/" -m pytest tests/ -vv &&
+        cd .. && cp server/.coverage . &&
```
nadzyah, line 32: this change was made according to this solution for fixing coverage report path: https://github.com/pytest-dev/pytest-cov/issues/305#issuecomment-508691008

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+                (
+                    models.Certificate.release_id == release.id
+                    if release is not None
+                    else true
+                ),
+            )
+        )
+        .distinct()
+        .all()
+    )
```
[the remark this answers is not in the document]
nadzyah: I've updated the function to require board to be specified. Also, I don't see any join with release table, or I'm missing something

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+        .join(
+            models.device_report_association,
+            models.Device.id == models.device_report_association.c.device_id,
+        )
+        .join(
+            models.Report,
+            models.device_report_association.c.report_id == models.Report.id,
+        )
+        .join(models.Certificate, models.Report.certificate_id == models.Certificate.id)
+        .filter(models.Certificate.machine_id.in_(machine_ids))
```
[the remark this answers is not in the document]
nadzyah: No, certs are still associated with machines
https://github.com/canonical/hexr/blob/03aec88ed4262eb9f998d11366a051ef1b54c2f9/apps/certificates/models.py#L102-L106

server/hwapi/endpoints/certification/logic.py
nadzyah: probably the naming is confising, I'll change it

server/hwapi/data_models/repository.py
plars: I had the same question! I thought I was being dense and I eventually convinced myself that it was likely the PCI-ID, USB-ID, etc. dependent on the device? I am curious if that's true now though, and agree it would be nice to make this more clear, either in the naming or in some documentation

server/hwapi/data_models/repository.py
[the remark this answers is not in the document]
omar-selo: Oh sorry I meant Certificate table not Release table

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+        db.query(models.Device)
+        .join(models.Vendor)
+        .filter(
+            and_(
+                models.Vendor.name.in_(
+                    [
+                        board_validator.manufacturer,
+                        clean_vendor_name(board_validator.manufacturer),
+                    ]
+                ),
```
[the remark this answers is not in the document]
nadzyah: thank you for the suggestion, I also thought about vendor aliases
however, this PR is already big enough, can we do it in another PR?

server/hwapi/data_models/repository.py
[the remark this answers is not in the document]
nadzyah: we join certificates here to get reports for the machines, since reports are related to machines through certs

server/hwapi/endpoints/certification/logic.py
nadzyah: this logic has been removed, mark as resolved

server/hwapi/data_models/repository.py
nadzyah: I've removed this logic, but in general "version" for CPU and GPU means "model name", that's why we compare it with a name

server/hwapi/data_models/repository.py
nadzyah: this logic has been removed

server/hwapi/data_models/repository.py
nadzyah: hm, since we don't store bus as an object, I don't see a reason to do it. shall we change it?

server/hwapi/data_models/repository.py
nadzyah: yeah, it's PCI or USB ID. We use the same naming convension in C3

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +111,253 @@ def get_or_create(
...
+            )
+        )
+        .distinct()
+        .all()
+    )
+
+
+def get_release_from_os(db: Session, os: OSValidator) -> models.Release:
+    """Return release object matching given OS data"""
```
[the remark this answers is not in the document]
nadzyah: I've updated the function to avoid confusions

server/hwapi/data_models/repository.py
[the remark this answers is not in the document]
omar-selo: that sounds fine with me

server/hwapi/data_models/repository.py
nadzyah: I've removed Query usage in the code

server/hwapi/data_models/repository.py
mz2: No, you don't need to reflect the type name in a variable or a property.

server/hwapi/data_models/repository.py
omar-selo: hmm, I can still see it, it's not a blocker though

nadzyah: I mean, I don't even import the Query logic anymore

server/hwapi/endpoints/certification/logic.py
```diff
@@ -46,27 +51,203 @@ def is_certified(system_info: CertificationStatusRequest, db: Session):
...
-            os=data_validators.OSValidator(
-                distributor="Canonical Ltd.",
-                version=latest_certificate.release.release,
-                codename=latest_certificate.release.codename,
-                kernel=data_validators.KernelPackageValidator(
-                    name=kernel.name, version=kernel.version, signature=kernel.signature
-                ),
-                loaded_modules=[],
+    return True, CertifiedResponse(
+        os=data_validators.OSValidator(
```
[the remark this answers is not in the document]
omar-selo: would it make sense to rename the validators or add new objects?

server/hwapi/data_models/repository.py
nadzyah: I've removed it here: 11c1d7a

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +61,163 @@ def get_or_create(
...
+    )
+
+
+def get_vendor_by_name(db: Session, name: str) -> models.Vendor | None:
+    """Find vendor by name (cleaned-up)"""
+    return (
+        db.query(models.Vendor)
+        .filter(models.Vendor.name.ilike(f"%{clean_vendor_name(name).lower()}%"))
+        .first()
+    )
```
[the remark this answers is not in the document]
nadzyah: here we use it for vendor name, and if a vendor contains its name substring (like the same name but without "Inc"), it's still the same vendor
For other cases we don't use it with `%` sign, so it means that we just make the search case-insensitive and there is no "contains" check

server/hwapi/endpoints/certification/logic.py
```diff
@@ -18,55 +18,36 @@
...
+) -> tuple[models.Device, models.Bios]:
+    board = repository.get_board(
+        db, board_data.manufacturer, board_data.product_name, board_data.version
     )
-    if configurations is None:
-        return False, None
+    bios = repository.get_bios(db, bios_data.vendor, bios_data.version)
+    if not board or not bios:
+        raise ValueError("Hardware not certified")
+    return board, bios
```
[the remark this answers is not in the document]
nadzyah: I thought about removing this function but without it the endpoint looks overwhelmed
I'll update the function name

server/hwapi/data_models/repository.py
```diff
@@ -107,3 +61,163 @@ def get_or_create(
...
+def get_board(
+    db: Session, vendor_name: str, product_name: str, version: str
+) -> models.Device | None:
+    """Return device object (category==BOARD) matching given board data"""
+    return (
+        db.query(models.Device)
+        .join(models.Vendor)
+        .filter(
+            and_(
+                models.Vendor.name.ilike(vendor_name),
```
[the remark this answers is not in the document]
nadzyah: with `get_vendor_by_name` we retrieve the machine vendor, not the device vendor
they are still the same class object, however machine vendor is not the same vendor object as machine device vendor

server/hwapi/data_models/repository.py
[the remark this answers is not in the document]
omar-selo: Oh ok, I hope it's fine for vendors then

server/hwapi/data_models/repository.py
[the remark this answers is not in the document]
omar-selo: oh yeah right

omar-selo (approved):
Looks good!

PR #139, opened 2024-05-17
Add coverage command and PR annotation

nadzyah:
This PR simply modifies the server GH action and adds coverage report to it

### DIFF
--- a/.github/workflows/test_server.yaml
+++ b/.github/workflows/test_server.yaml
@@ -44,6 +44,33 @@ jobs:
       - name: Run tests with docker-compose
         run: |
           docker-compose up --attach-dependencies --abort-on-container-exit hwapi-test
+          sudo chown $USER -R coverage
+          sudo chgrp $USER -R coverage
+      - name: Upload coverage.xml as an artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: test_coverage
+          path: ./server/coverage/coverage.xml
+          if-no-files-found: error
+          retention-days: 2
+
+  report-coverage:
+    runs-on: [self-hosted, linux, large, jammy, x64]
+    if: github.event_name == 'pull_request'
+    needs: [test]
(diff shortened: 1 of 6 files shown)


.github/workflows/test_server.yaml
```diff
@@ -44,6 +44,33 @@ jobs:
       - name: Run tests with docker-compose
         run: |
           docker-compose up --attach-dependencies --abort-on-container-exit hwapi-test
+          sudo chown $USER -R coverage
+          sudo chgrp $USER -R coverage
+      - name: Upload coverage.xml as an artifact
+        uses: actions/upload-artifact@v3
```
omar-selo, line 50: any reason not to use v4?

.github/workflows/test_server.yaml
```diff
@@ -44,6 +44,33 @@ jobs:
...
+          if-no-files-found: error
+          retention-days: 2
+
+  report-coverage:
+    runs-on: [self-hosted, linux, large, jammy, x64]
+    if: github.event_name == 'pull_request'
+    needs: [test]
+    steps:
+      - name: Download build tests coverage report
+        uses: actions/download-artifact@v3
```
omar-selo, line 63: same here

omar-selo (commented):
Looks good just a couple of comments

scripts/generate_openapi_html.sh
```diff
@@ -5,6 +5,7 @@ if git diff --cached --name-only | grep --quiet "server"
 then
     cd server/schemas/
 
+    which npx
```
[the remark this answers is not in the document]
nadzyah: The problem is that the hook passes even if npx is not installed, that's why I've decided to fix it here and add check that we have this command. I've adjusted the script to exit if it's not found

omar-selo (approved):
LGTM!

PR #149, opened 2024-05-30
Rebuild the OCI and Charm periodically

plars:
Not sure if we want this just yet, but it might be good to start testing it and I don't see a downside to doing it now. This will periodically rebuild the hardware-api OCI image and charm on a schedule (weekly for now, can increase later easily if we want).  When rebuilding it, we should also get the latest c3 data in the sqlite db

### DIFF
--- a/.github/workflows/publish_server.yml
+++ b/.github/workflows/publish_server.yml
@@ -5,6 +5,8 @@ on:
       - "server/**"
       - ".github/workflows/publish_server.yml"
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


nadzyah (approved):

nadzyah (approved):
Ah, didn't notice that you actually rebuild the image in GH, sorry 😬 
I thought it was just charm rebuild part. Sorry again, I obviously approve this change and thank you for it

PR #153, opened 2024-06-10
Remove HTML schema from repo

nadzyah:
This PR just removed the `openapi.html` schema from the repo, since there is no usage of it (GH doesn't have HTML preview, and we have the same view generated on github pages: https://canonical.github.io/hardware-api)


mz2 (approved):
LGTM

PR #154, opened 2024-06-13
Exclude debian/ dir from publishing to crates and vendor dependencies

nadzyah:
This PR excludes the `debian/` dir that we store on github from publishing on crates.
Since it doesn't affect the rust package itself, we shouldn't have it on crates.io

It also updates the dependencies and vendors them


Hook25 (dismissed):

omar-selo:
Records have been broken 
image

PR #155, opened 2024-06-14
Prepare hwlib for publishing to crates

nadzyah:
This PR prepares the package to be published on crates.io, since we need it before packing it as debian

The PR does the following:
- removes `debian/` dir from tracking it, so we don't publish it on crates
- vendor the dependencies using cargo-vendor-filterer (all non-linux dependencies got replaced with stub)
- updates the test_client action to perform the offline package installation (with `cargo build --offline`, preferably on focal)

One of the main reasons why we use the vendor approach is that it's the [ubuntu-mir policy](https://github.com/canonical/ubuntu-mir/tree/1d0331b99c6ad60482b40925e9e00bd906399b18) for rust libs
```
RULE: - The rust ecosystem currently isn't yet considered stable enough for
RULE:   classic lib dependencies and transitions in main; therefore the
RULE:   expectation for those packages is to vendor (and own/test) all
RULE:   dependencies (except those provided by the rust runtime itself).
```


nadzyah:
These are the only files that need to be reviewed. All other once (inside the `vendor/` dir) are the dependencies that I haven't modified
image

plars (commented):
I was able to build the source package following your directions. However, I had trouble building the binary package locally at least. It keeps getting stuck on pyo3 and I'm not sure why. It's complaining that the version it can find is 0.20.2 (which is the one from the deb dependency) and not 0.21.2 which is specified in the Cargo.toml.  It's highly likely that I'm just doing something wrong here, but I was curious if you did a dput to a ppa and got it to build the binary package somewhere?

I was also wondering if the README is still correct that we should target mantic. It seems like we should target noble, or more likely oracular?
Another thing I don't see documented is how to update the vendored packages with `cargo vendor`. This might be good to have in the README in case anyone needs to update them later.

Finally, I was wondering if anyone showed you other examples of rust packages that used vendored libraries. I'm curious if it's common to put everything in one place, or whether it makes more sense to run keep them in another repo or possibly run the `cargo vendor` command when doing the build?

nadzyah:
> I was able to build the source package following your directions. However, I had trouble building the binary package locally at least. It keeps getting stuck on pyo3 and I'm not sure why. It's complaining that the version it can find is 0.20.2 (which is the one from the deb dependency) and not 0.21.2 which is specified in the Cargo.toml.  It's highly likely that I'm just doing something wrong here, but I was curious if you did a dput to a ppa and got it to build the binary package somewhere?

We shouldn’t use deb dependencies at all. You need to specify the vendor directory in the `~/.cargo/config.toml`directory, so cargo uses the vendor directory for all dependencies

> I was also wondering if the README is still correct that we should target mantic. It seems like we should target noble, or more likely oracular?

I’ll update it when in next PR where I pack the lib as deb


> Another thing I don't see documented is how to update the vendored packages with `cargo vendor`. This might be good to have in the README in case anyone needs to update them later.

I was thinking about doing it in a follow-up PR since it’s hard to check it here. The CI also checks that the package is installable offline with the vendored deps

> Finally, I was wondering if anyone showed you other examples of rust packages that used vendored libraries. I'm curious if it's common to put everything in one place, or whether it makes more sense to run keep them in another repo or possibly run the `cargo vendor` command when doing the build?

I'll try to find some examples

plars (commented):
One thing I found when building the binary package locally is that I get this error:
```
# Install the Python bindings generated by maturin
python3 -m pip install ./*.whl --no-deps --target debian/hwlib/usr/lib/python3/dist-packages/ --no-cache-dir
WARNING: Requirement './*.whl' looks like a filename, but the file does not exist
ERROR: *.whl is not a valid wheel filename.
```
I had done the `maturin build --release ...` but it places the wheel file in a different directory. However, even if I specify this as the output directory, it looks like the debian build process cleans it before starting. I'm not sure if this would also affect source builds on launchpad, but at the very least, maturin by default seems to put it in `(project_root)/target/wheels` which means it won't be in the orig tarball that gets shipped with the source package.

Are you seeing this also? I was trying to force the maturin build to happen during override_dh_auto_build and it almost worked, but I'm getting an error later in the process:
```
readelf: Error: no .dynamic section in the dynamic segment
objcopy: warning: debian/hwlib/usr/lib/python3/dist-packages/hwlib/hwlib.cpython-312-x86_64-linux-gnu.so has a corrupt string table index
objcopy: unable to modify 'debian/hwlib/usr/lib/python3/dist-packages/hwlib/hwlib.cpython-312-x86_64-linux-gnu.so' due to errors
```
(this could also be due to trying to use the python3-maturin package for this part of it rather than the pip-installed version.
I'm still trying some things, but wanted to share this and see if you were seeing something similar when trying to build it

nadzyah:
> (this could also be due to trying to use the python3-maturin package for this part of it rather than the pip-installed version.

I've also faced this error, and it's recommended to uses pip-installed version, because it required `patchelf` dependency (or you can install it with apt). It should fix the problem

However, first, we need to make sure that the rust library can be built correctly because we pack it as Debian first. Then we'll work on packing the Python bindings as a Python library. 

Also, the workflow for this will be changed. This is because the maturing command we’re using generates the binary .so file (you’re getting an error with) and debiab policy forbids adding binaries to archive source.

plars:
> Also, the workflow for this will be changed. This is because the maturing command we’re using generates the binary .so file (you’re getting an error with) and debiab policy forbids adding binaries to archive source.

Yes, but the current state of the package is that it's expecting the wheel to be at that path and it won't be. So I don't think the package can completely build at all right now because it will break on that step. If the intent was just to get to a point where you have the rust part of the build working inside the packaging, then that's ok. I guess this is just a step towards the functional deb?

nadzyah:
> Yes, but the current state of the package is that it's expecting the wheel to be at that path and it won't be. So I don't think the package can completely build at all right now because it will break on that step. If the intent was just to get to a point where you have the rust part of the build working inside the packaging, then that's ok. I guess this is just a step towards the functional deb?

I have another branch where I’ve updated the debiab dir and build the package. We’ve completely changed the approach, we’re building the rust and python libs separately. Currently we focus on the rust part
Also, this PR just prepares the package to be published on crates. The instructions for preparing it to be published as debian are wrong here, and I’m taking care of it in the `update-client-debian` branch

nadzyah:
@plars 

> If the intent was just to get to a point where you have the rust part of the build working inside the packaging, then that's ok.

Yeah, this is what we’re doing as a new approach! First, we build the rust part, and then we build the python part relying on the rust one

You can find more details on this here: https://docs.google.com/document/d/1lGyr2pbKz5GgIwAy5chwb3WT0p1nZN2mbdRgXQGtNXw/edit

plars (approved):
Thanks for the link to that document, that puts it in better context. I think I was at least able to reproduce the parts that are expected to work now from the README and offline builds are working well with the vendor directory. +1

PR #169, opened 2024-07-04
Update debian directory for hwlib

nadzyah:
This PR closes [C3-682](https://warthogs.atlassian.net/browse/C3-682)

It updates the `client/hwlib/debian/` directory to be used for publishing the `librust-hwlib-dev` package.
The changelog file prepares it to be published under the `0.0.1` version for the focal distro.

Also, here the CI job has been created for testing how the package can be built with the `sbuild` tool

[C3-682]: https://warthogs.atlassian.net/browse/C3-682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/test_client.yaml
```diff
@@ -34,7 +34,10 @@ jobs:
         run: |
           cd client/hwlib/
           ./debian/vendor-rust.sh
-          echo -e '[source.crates-io]\nreplace-with = "vendored-sources"\n\n[source.vendored-sources]\ndirectory = "'"$PWD"'/vendor"' > ~/.cargo/config.toml
+          echo '[source.crates-io]' >>  ~/.cargo/config.toml
```
zyga, line 37: This looks like a here-doc
cat >>~/.cargo/config.toml << __CONFIG__
[source.creates-io]
replace-with = "vendored-sources"
[source.vendored-sources]
directory = "$PWD/vendor"
__CONFIG__

.github/workflows/test_hwlib_debian_build.yaml
```diff
@@ -0,0 +1,80 @@
...
+    runs-on: [self-hosted, linux, large, noble, x64]
+    defaults:
+      run:
+        working-directory: client/hwlib
+    steps:
+    - name: Checkout the repository
+      uses: actions/checkout@v4
+    - name: Download dependencies
+      run: |
+        sudo apt update
```
zyga, line 30: Add `set -xeu` to make sure that you:
- see what is executed with trace (x)
- fail if anything fails in a simple statement (e)
- fail if there's an undefined variable being referenced (u)

.github/workflows/test_hwlib_debian_build.yaml
```diff
@@ -0,0 +1,80 @@
...
+      run: |
+        mkdir -p ~/.cache/sbuild
+        echo
+        mmdebstrap --variant=buildd --components=main,restricted,universe focal ~/.cache/sbuild/focal-amd64.tar.zst
+
+        echo "Running sbuild-debian-developer-setup"
+        sudo sbuild-debian-developer-setup
+        sudo sbuild-adduser $USER
+        newgrp sbuild
+        echo "Creating ~/.sbuildrc"
```
zyga, line 66: Another here-doc, please rewrite that for simplicity.

Tip: here-docs have two forms __FOO__ and '__FOO__' which differ in expansion semantics, use the right one to avoid lots of escape characters.

client/README.md
```diff
@@ -70,44 +72,99 @@ Now you can use the lib in your Python code:
...
+
+Not you can build the binary itself:
+
+```bash
+sbuild /path/to/.dsc -d focal
+```
+
+After that, you can publish the package by rinning:
+
+```bash
```
zyga, line 168: Nitpick, this is just plain `sh`

client/hwlib/debian/rules
```diff
@@ -1,47 +1,9 @@
...
-	install -Dm755 target/$(DEB_HOST_RUST_TYPE)/release/libhwlib.so debian/hwlib/usr/lib/$(DEB_HOST_MULTIARCH)/
-	# Install the Python bindings generated by maturin
-	python3 -m pip install ./*.whl --no-deps --target debian/hwlib/usr/lib/python3/dist-packages/ --no-cache-dir
-
-override_dh_shlibdeps:
-	dh_shlibdeps -l$(CURDIR)/debian/hwlib/usr/lib/python3/dist-packages/hwlib.libs
-
-override_dh_dwz:
+override_dh_auto_test:
+	/usr/share/cargo/bin/cargo test -- --test-threads=1
```
zyga, line 9: Nice reduction!

Why do we need --test-threads=1? Please add a comment for the next maintainer.

client/hwlib/debian/vendor-rust.sh
```diff
@@ -7,7 +8,16 @@ if [ ! $(which cargo-vendor-filterer) ]; then
...
 fi
 
+cargo vendor-filterer "${CARGO_VENDOR_DIR:-vendor}"
+
 # Some crates are shipped with .a files, which get removed by the helpers during the package build as a safety measure.
 # This results in cargo failing to compile, since the files (which are listed in the checksums) are not there anymore.
-# For those crates, we need to replace their checksum with a more general one that only lists the crate checksum, instead of each file.
-cargo vendor-filterer "${CARGO_VENDOR_DIR:-vendor}"
+# For those crates, we need to remove their checksum
+find "${CARGO_VENDOR_DIR:-vendor}" -name ".cargo-checksum.json" -exec bash -c '
```
zyga, line 16: This is somewhat hard to read. Peharps something more along the lines:
```sh
for json_file in $(find ...); do
   ...
done
```

Make sure to shellcheck the logic.

zyga (commented):
Quick pass, most of this looks good. I left a few comments on the shell pieces. Have a look please.

client/hwlib/debian/rules
nadzyah: This is how we run the tests in general for hwlib even in CI.
We do it this way because our tests (for now, will be changed with [this task](https://warthogs.atlassian.net/browse/C3-681)) rely on the environment variable value. So, `cargo test -- --test-threads=1` forces the tests to run sequentially, which avoids potential issues with environment variables being modified by concurrent tests

.github/workflows/test_hwlib_debian_build.yaml
nadzyah: I've updated the code

.github/workflows/test_hwlib_debian_build.yaml
nadzyah: I've added the `set -xeu` to steps with multiple commands

client/hwlib/debian/vendor-rust.sh
nadzyah: I've tried to simplify the logic with your suggestion

.github/workflows/test_hwlib_debian_build.yaml
```diff
@@ -0,0 +1,85 @@
...
+        ./debian/vendor-rust.sh
+    - name: Generate checksums
+      run: ./debian/generate_checksums.py
+    - run: dpkg-buildpackage -us -uc -S
+    - name: Setup sbuild and build the package
+      run: |
+        set -xeu
+        mkdir -p ~/.cache/sbuild
+        echo
+        mmdebstrap --variant=buildd --components=main,restricted,universe focal ~/.cache/sbuild/focal-amd64.tar.zst
```
zyga, line 63: You can use an action to cache this. Consider it follow up material.

client/hwlib/debian/generate_checksums.py
```diff
@@ -0,0 +1,43 @@
...
+        while chunk := f.read(8192):
+            sha256.update(chunk)
+    return sha256.hexdigest()
+
+def generate_checksums(root_dir):
+    """Generate a dictionary of file paths and their SHA-256 checksums."""
+    checksums = {}
+    for dirpath, dirnames, filenames in os.walk(root_dir):
+        # Exclude debian/ and vendor/ directories
+        if 'debian' in dirnames:
```
zyga, line 20: This is okay for now but should only match in the specific two locations. As coded it matches any directory with those names.

client/hwlib/debian/rules
zyga: You just said: the code is buggy.

Please add this comment to the repo code, file a ticket and link it there so that it can be removed.

Tests that modify environment variables need to grab a lock or use some other method. A single threaded runner is not a workaround.

zyga (approved):
LGTM but please see my comments inline.

client/hwlib/debian/generate_checksums.py
nadzyah: It should match any files in the directory excluding debian/ and vendor/

zyga (approved):
Looks good,  thanks for the improvement to root dir matching

PR #173, opened 2024-07-08
Add match against bios firmware_revision

nadzyah:
Since Bios object is one of the key components that we use in the logic to match against a machine in the DB, we need to get the exact Bios object that the machine uses. So, we need to be precise when we look for a Bios object, that's why I'm adding here `firmware_revision` as another condition for matching against Bios object


andrejvelichkovski (approved):
LGTM

PR #175, opened 2024-07-10
Update changelog and related files to publishing for oracular

nadzyah:
This PR updates the changelog, CI job, and readme to use oracular distro instead of focal.
The current PPA where the lib is published will be wiped up, so that's why the changelog is modified instead of being updates

For noble, jammy, and focal, we'll use backporting. Here you can see an example of how I've tested it in a test PPA: https://launchpad.net/~nhutsko/+archive/ubuntu/ppa/+packages?field.name_filter=rust&field.status_filter=published&field.series_filter=

Also, I've updated the `vendor_rust.sh` script to exclude the shell scripts from the `vendor/vcpkg` libraries. We don't use these scripts in our code and don't depend on it, but they generate lintian errors because of the incorrect bash syntax

### DIFF
--- a/client/hwlib/debian/vendor-rust.sh
+++ b/client/hwlib/debian/vendor-rust.sh
@@ -15,14 +15,20 @@ cargo vendor-filterer "$CARGO_VENDOR_DIR"
 # Delete .sh files in the vendor/vcpkg directory if it exists
 if [ -d "$CARGO_VENDOR_DIR/vcpkg" ]; then
     find "$CARGO_VENDOR_DIR/vcpkg" -name "*.sh" -exec rm -f {} +
+    # Remove checksums for the deleted *.sh files that are not used by the code but generate lintian errors
+    for json_file in $(find "${CARGO_VENDOR_DIR}/vcpkg" -name ".cargo-checksum.json"); do
+        tmp_file=$(mktemp)
+        jq '.files |= with_entries(select(.key | endswith(".sh") | not))' "$json_file" > "$tmp_file"
+        mv "$tmp_file" "$json_file"
+    done
 fi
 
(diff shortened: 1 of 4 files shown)


client/hwlib/debian/vendor-rust.sh
```diff
@@ -12,6 +12,17 @@ fi
 
 cargo vendor-filterer "$CARGO_VENDOR_DIR"
 
+# Delete .sh files in the vendor/vcpkg directory if it exists
+if [ -d "$CARGO_VENDOR_DIR/vcpkg" ]; then
+    find "$CARGO_VENDOR_DIR/vcpkg" -name "*.sh" -exec rm -f {} +
```
zyga, line 17: What is the trailing plus for? Was this supposed to be \; ?

nadzyah: I guess it doesn't matter much in this case. The `+` sign list files in a single line, and the `;` sign lists each new file with a new line

```bash
$ find vendor/vcpkg/ -name "*.sh" -exec ls -f {} +
vendor/vcpkg/tests/run.sh  vendor/vcpkg/tests/vcpkgrs_target.sh  vendor/vcpkg/setup_vcp.sh

$ find vendor/vcpkg/ -name "*.sh" -exec ls -f {} \;
vendor/vcpkg/tests/run.sh
vendor/vcpkg/tests/vcpkgrs_target.sh
vendor/vcpkg/setup_vcp.sh
````

However, both of the commands remove the `*.sh` files:
image

Anyway, I can change it to `\;` if you want

zyga: Nah, it's okay. I was just unaware of + mode.

PR #176, opened 2024-07-11
Update rust deps

nadzyah:
This PR simply updates the Rust dependencies in the project and adjust the code to remove the usage of deprecated functions

Also, here we add Cargo.lock file to the `hwlib` dir to include it to the debian package, as required by MIR policy


client/hwlib/Cargo.toml
```diff
@@ -20,12 +20,12 @@ name = "hwlib"
 crate-type = [ "cdylib", "rlib",]
 
 [target.'cfg(all(target_os = "linux"))'.dependencies]
-serde_json = "1.0.41"
+serde_json = "1.0.0"
```
p-gentili, line 23: This feels a bit loose, let's do "~1.0" instead.

client/hwlib/Cargo.toml
```diff
@@ -20,12 +20,12 @@ name = "hwlib"
...
 once_cell = "1.19.0"
 reqwest = { version = "0.12.0", features = [ "json",] }
-serde = { version = "1.0.104", features = [ "derive",] }
+serde = { version = "1.0.0", features = [ "derive",] }
```
p-gentili, line 26: Same as above.

p-gentili (changes requested):

p-gentili (approved):
LGTM, thanks!

PR #182, opened 2024-07-19
Allow bios field to be optional

nadzyah:
Since some devices don't have bios (like Raspberry Pi) and we store the value for Bios field in C3 as None, here we allow the bios field to be optional
It means that the bios field can be specified as None in certification requests and responses.

### DIFF
--- a/server/hwapi/endpoints/certification/logic.py
+++ b/server/hwapi/endpoints/certification/logic.py
@@ -26,8 +26,8 @@ from hwapi.data_models.data_validators import BoardValidator, BiosValidator
 
 
 def find_main_hardware_components(
-    db: Session, board_data: BoardValidator, bios_data: BiosValidator
-) -> tuple[models.Device, models.Bios]:
+    db: Session, board_data: BoardValidator, bios_data: BiosValidator | None
+) -> tuple[models.Device, models.Bios | None]:
     """
     A function to get "main hardware components" like board and bios. Can be extended
     in future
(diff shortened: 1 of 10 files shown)


server/tests/endpoints/certification/test_certification.py
```diff
@@ -429,3 +429,93 @@ def test_all_criteria_matched(generator: DataGenerator, test_client: TestClient)
...
+                "distributor": "Canonical Ltd.",
+                "version": release.release,
+                "codename": release.codename,
+                "kernel": {"name": None, "version": "5.7.1-generic", "signature": None},
+            },
+            "processor": {
+                "codename": processor.codename,
+                "frequency": 2.0,
+                "manufacturer": "Intel Corp.",
+                "version": "Intel(R) Core(TM) i7-7600U CPU @ 2.40GHz",
```
omar-selo, line 492: Just a question. I see here that the processor version `ntel(R) Core(TM) i7-7600U CPU @ 2.40GHz` is different from the processor name specified above `Intel(R) Core(TM) i5-7300U CPU @ 2.60GHz`. Yet the resulting status is `Certified`. Does that mean that the processor doesn't need to be exactly the same to be considered certified?

server/hwapi/endpoints/certification/logic.py
```diff
@@ -26,25 +26,27 @@
...
-    )
-    if not board or not bios:
+    if not board:
         raise ValueError("Hardware not certified")
-    return board, bios
+    if bios_data:
+        bios = repository.get_bios(
+            db, bios_data.vendor, bios_data.version, bios_data.firmware_revision
+        )
+        return board, bios
```
omar-selo, line 44: if `bios_data` is not `None` shouldn't we still check that we found it in the repository and otherwise raise `ValueError("Hardware not certified")`

omar-selo (commented):
Looks good, just a couple of comments

server/tests/endpoints/certification/test_certification.py
nadzyah: we match processor codename, not the exact model name. So, it means that the processor version (model name) can be different, but codename must be the same
you can check it here: https://github.com/canonical/hardware-api/blob/1bb344/server/hwapi/endpoints/certification/logic.py#L55-L59

server/hwapi/endpoints/certification/logic.py
nadzyah: you're right about that, I'll check it

server/tests/endpoints/certification/test_certification.py
omar-selo: Got it, thanks for the clarification

omar-selo (approved):
Looks good!

PR #183, opened 2024-07-19
Import cpu_codename field for processor devices

nadzyah:
Since now we have the proper CPU codename in the C3 /api/v2/public-devices/ endpoint, we update the devices with type PROCESSOR to use this codename as the Device.codename field

Also, this PR speeds up the import from C3 by using threading and submitting the request to next_url while processing the received data

### DIFF
--- a/server/hwapi/external/c3/client.py
+++ b/server/hwapi/external/c3/client.py
@@ -20,7 +20,9 @@
 
 import requests
 import logging
-from typing import Callable, Type
+import threading
+import queue
+from typing import Callable, Type, Any
 from pydantic import BaseModel
 
 from sqlite3 import IntegrityError as SQLite3IntegrityError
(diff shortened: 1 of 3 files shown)


server/hwapi/external/c3/client.py
```diff
@@ -250,4 +280,4 @@ def _load_devices_from_response(
         )
         if created or device not in report.devices:
             report.devices.append(device)
-            self.db.commit()
+        self.db.commit()
```
nadzyah, line 283: I'm not sure whether I should commit everything at the end or commit separately after changing the codename, and then after modifying the report

omar-selo (approved):
Looks good. Just a suggestion, feel free to take it or leave it. Threading is a bit old school in python. Asyncio is easier to use, easier to read, and from my experience more performant

nadzyah:
> Looks good. Just a suggestion, feel free to take it or leave it. Threading is a bit old school in python. Asyncio is easier to use, easier to read, and from my experience more performant

Thanks for your suggestion! I guess I'd like to refactor it later since it would take much time for now

PR #190, opened 2024-07-26
Drop threading in the c3 client logic

nadzyah:
It turned out that C3 cannot handle so many frequent requests in a row and starts returning 500 error, I've decided to drop the threading solution and revert the script logic

Also, to avoid printing the progress bar with the same output multiple times, I've updated the script to print it only when the percentage changes, and change it to print only one digit after the decimal point.

### DIFF
--- a/server/hwapi/external/c3/client.py
+++ b/server/hwapi/external/c3/client.py
@@ -20,8 +20,6 @@
 
 import requests
 import logging
-import threading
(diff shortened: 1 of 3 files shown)


omar-selo (approved):
Looks good, I guess we accidentally DDOSed C3 :sweat_smile:

PR #200, opened 2024-08-23
Move CPU ID and codename matching logic to server

nadzyah:
This PR refactors the handling of CPU ID matching by moving the logic from the client side to the server side. Instead of sending the Processor codename, the client now sends the raw CPU ID bytes directly to the server. This change simplifies client-side logic and enables more flexible and centralized management of the CPU ID map.

## Key Changes
1. CPU ID Matching Logic Moved to Server:
    - The client now sends CPU ID bytes to the server, which performs the matching against the processor codename. This allows the CPU ID map to be updated on the server side without requiring client updates.
2. Centralized CPU ID Map Retrieval:
    - To avoid duplicating the CPU ID map within the server codebase, this PR introduces logic to retrieve CPU IDs and codenames directly from the C3 service.
    - The server fetches this data from the new C3 endpoint (referenced in https://github.com/canonical/hexr/pull/388) during the build process. This approach minimises runtime overhead and ensures the CPU ID map is always up to date.
3. Data Storage:
    - Retrieved CPU IDs and their corresponding codenames are stored in the `CpuId` table. This ensures that the matching logic can efficiently operate on locally stored data, reducing the need for repeated API calls.

## Benefits
- Reduced Client Complexity: By shifting the matching logic to the server, client-side complexity is reduced, and future updates to the CPU ID map no longer require client-side changes.
- Centralized Management: The CPU ID map is now managed centrally through the C3 service, ensuring consistency across all clients without the need for frequent updates.
- Efficiency: The server fetches and stores CPU ID data at build time, eliminating the need for real-time fetching during each match operation, leading to improved performance.


omar-selo (dismissed):
At least the python side of things looks fine

nadzyah:
@TheSignPainter98 @sminez would appreciate your review of the client part.

client/hwlib/src/collectors/hardware_info.rs
```diff
@@ -78,10 +75,12 @@ impl TryFrom<(&SMBiosProcessorInformation<'_>, &Path)> for Processor {
...
     fn try_from(value: (&SMBiosProcessorInformation, &Path)) -> Result<Self> {
         let (processor_info, max_cpu_frequency_filepath) = value;
-        let cpu_id = CpuId::from_smbios(processor_info)?;
+        let cpu_id = processor_info
+            .processor_id()
+            .ok_or_else(|| anyhow!("processor ID not found"))?;
         let cpu_freq = CpuFrequency::from_k_hz_file(max_cpu_frequency_filepath)?.get_m_hz();
         Ok(Processor {
-            codename: cpu_id.codename().unwrap_or_else(|| "Unknown".to_string()),
+            identifier: *cpu_id,
```
TheSignPainter98, line 83: Not sure if it's something for this PR, but there's a minor inconsistency with the naming here as we have both `id` and `identifier` to refer to the same concept. If possible, it would be nice to stick to just one, although I don't know which would be easiest, or indeed whether this is too much of a hassle

client/hwlib/src/collectors/hardware_info.rs
```diff
@@ -265,7 +264,10 @@ mod tests {
             get_test_filepath("cpuinfo_max_freq").as_path(),
         ))
         .unwrap();
-        assert_eq!(processor.codename, "Tiger Lake");
+        assert_eq!(
+            processor.identifier,
+            [0xc1, 0x6, 0x8, 0x0, 0xff, 0xfb, 0xeb, 0xbf]
```
TheSignPainter98, line 269: Just to avoid magic constants, I think it could be nice to add a comment here noting that this is the Tiger Lake ID. (I assume this is the Tiger Lake ID)

client/hwlib/src/collectors/hardware_info.rs
```diff
@@ -78,10 +75,12 @@ impl TryFrom<(&SMBiosProcessorInformation<'_>, &Path)> for Processor {
 
     fn try_from(value: (&SMBiosProcessorInformation, &Path)) -> Result<Self> {
         let (processor_info, max_cpu_frequency_filepath) = value;
-        let cpu_id = CpuId::from_smbios(processor_info)?;
+        let cpu_id = processor_info
+            .processor_id()
+            .ok_or_else(|| anyhow!("processor ID not found"))?;
```
TheSignPainter98, line 80: This may not be the case, but I think there's a slightly strange interaction here. I'm assuming that having `[0; 8]` means 'unknown', but here we are getting an error that the processor ID hasn't been found (somehow these seem similar to me). I think 'not found' may be slightly ambiguous here: is this not found as in not found locally, or not found as in found locally but not recognised?

I think the issue may be that `.processor_id` returns an `Option` and the `identifier` field has behaviour similar to an `Option`, but if these were given the same type, it would be easy for someone to accidentally come along and put `identifier: processor_info.processor_id()` and miss out the important case of the processor ID missing. I think this could be fixed by adding a new enum to represent whether the ID is known/unknown, which specifically _doesn't_ implement `From<Option<_>>`. (This would then force a manual conversion and hence for the locally-missing case to be explicitly considered)

This may be overkill though, but I just have a slightly odd feeling

TheSignPainter98 (changes requested):
The Rust side of things is looking good! Its overall form seems nice to me, here are a couple of fine-tuning notes

client/hwlib/src/collectors/hardware_info.rs
```diff
@@ -98,7 +97,7 @@ impl TryFrom<(&Path, &Path)> for Processor {
         let cpu_info = CpuInfo::from_file(cpuinfo_filepath)?;
         let cpu_freq = CpuFrequency::from_k_hz_file(max_cpu_frequency_filepath)?.get_m_hz();
         Ok(Processor {
-            codename: String::new(),
+            identifier: [0; 8],
```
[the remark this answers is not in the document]
nadzyah: Updated in a43dfae

client/hwlib/src/collectors/hardware_info.rs
nadzyah: Updated in bb02da2

client/hwlib/src/collectors/hardware_info.rs
nadzyah: Added in bb02da2

client/hwlib/src/collectors/hardware_info.rs
nadzyah: Updated in bb02da2

nadzyah:
@omar-selo I've added the logic to allow `Processor identifier` to be None, as @TheSignPainter98 requested. Can you please take a look at the server changes?

server/hwapi/data_models/models.py
```diff
@@ -185,3 +185,9 @@ class Device(Base):
     reports = relationship(
         "Report", secondary=device_report_association, back_populates="devices"
     )
+
+
+class CpuId(Base):
+    __tablename__ = "cpu_id"
+    id_pattern: Mapped[str] = mapped_column(nullable=False, index=True)
```
omar-selo, line 192: I see in `repository.get_cpu_id_object` that you get all CpuIds then look for the appropriate one in python (not through SQL). So if we're not using the index on `id_pattern` it would be better to remove it

omar-selo (commented):
Looks good just a small suggestion

omar-selo (approved):
LGTM!

PR #216, opened 2024-10-16
Add tests and command wrapper for `os_info.rs` in `collectors` module

nadzyah:
This PR addresses part of [C3-883](https://warthogs.atlassian.net/browse/C3-883) by adding unit tests to the `os_info.rs` component of the `collectors` module. The key changes include:

A new `CommandRunner` trait is introduced to wrap the `std::process::Command::run` function. This abstraction enables better testability by allowing us to mock system command invocations.

In the tests, the `MockCommandRunner` is used to simulate command outputs, allowing us to verify behaviour for different system states. This mock implementation replaces the real command execution to make testing deterministic and isolated.

[C3-883]: https://warthogs.atlassian.net/browse/C3-883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/hwlib/src/collectors/os_info.rs
```diff
@@ -68,47 +81,90 @@ impl TryFrom<&Path> for KernelPackage {
...
+    use super::*;
+    use crate::helpers::test_utils::get_test_filepath;
+    use mockall::mock;
+
+    mock! {
+        CommandRunner {}
+        impl CommandRunner for CommandRunner {
+            fn run_command<'a>(&self, cmd: &str, args: &[&'a str]) -> Result<String>;
+        }
+    }
```
sminez, line 129: Rather than bringing in a mocking crate for this I'd be tempted to just hand write the impl you need for your test cases as the logic is fairly simple:

```rust
struct MockRunner<'a> {
    calls: HashMap<(&'a str, Vec<&'a str>), Result<&'a str>>,
}

impl<'a> MockRunner<'a> {
    fn new(calls: Vec<((&'a str, &[&'a str]), Result<&'a str>)>) -> Self {
        let calls = calls
            .into_iter()
            .map(|((cmd, args), res)| ((cmd, args.to_vec()), res))
            .collect();

        Self { calls }
    }
}

impl<'a> CommandRunner for MockRunner<'a> {
    fn run_command(&self, cmd: &str, args: &[&str]) -> Result<String> {
        match self.calls.get(&(cmd, args.to_vec())) {
            Some(res) => res.clone().map(String::from),
            None => panic!("missing mock: cmd={cmd:?} args={args:?}"),
        }
    }
}
```

I've put this in a [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c24709d65ab934711d094df1378362d4) for you to take a look at if you want to try it out :slightly_smiling_face:

client/hwlib/src/models/request_validators.rs
```diff
@@ -77,20 +77,25 @@ pub struct CertificationStatusRequest {
 }
 
 impl CertificationStatusRequest {
-    pub fn new(paths: Paths) -> Result<Self> {
+    pub fn new(paths: Paths, runner: Option<&dyn CommandRunner>) -> Result<Self> {
```
sminez, line 80: Passing an `Option` here and then using `unwrap_or` inside of the function as a way of having a default value is a little odd. The API will be easier to understand at the call site if you always require passing the `CommandRunner` impl, or alternatively you could keep `new` with the signature it had before and change it to a wrapper for `new_with_runner`:
```rust
impl CertificationStatusRequest {
    pub fn new(paths: Paths) -> Result<Self> {
        Self::new_with_runner(paths: Paths, runner: &RealCommandRunner)
    }

    pub fn new_with_runner(paths: Paths, runner: &dyn CommandRunner) -> Result<Self> {
        // Current impl of "new" minus the call to unwrap_or
    }
}
```

sminez (changes requested):
Looking good on the testing front but there are a couple of things you could change around the API structure to make things more idiomatic and easier to work with / maintain I think. :slightly_smiling_face:

TheSignPainter98 (changes requested):
This is great work, here are a couple of things to consider but on top of Innes's suggestions, this is mostly just polish

client/hwlib/src/collectors/os_info.rs
```diff
@@ -18,21 +18,35 @@
...
+
+impl CommandRunner for RealCommandRunner {
+    fn run_command(&self, cmd: &str, args: &[&str]) -> Result<String> {
+        let output = Command::new(cmd).args(args).output()?;
+        let stdout = String::from_utf8(output.stdout)?;
+        Ok(stdout)
+    }
+}
+
+impl TryFrom<(&Path, &dyn CommandRunner)> for OS {
```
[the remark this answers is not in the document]
nadzyah: I've updated the code, removed `dyn` (as indicated in the commit message: f497200) and replaced it with `impl`

client/hwlib/src/models/request_validators.rs
nadzyah: Updated in f497200

client/hwlib/src/collectors/os_info.rs
```diff
@@ -18,21 +18,35 @@
...
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, Result};
 use lazy_static::lazy_static;
 use regex::Regex;
 use std::{fs::read_to_string, path::Path, process::Command};
 
 use crate::models::software::{KernelPackage, OS};
 
-impl TryFrom<&Path> for OS {
+pub trait CommandRunner {
```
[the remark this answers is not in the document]
nadzyah: This is an interesting question. The `CertificationStatusRequest` struct is public, because it's used to prepare the request and collect the hardware data. It has the `pub` method `new`, which accepts only `&Paths` to the files that we can use to collect the data
Since OS module needs `CommandRunner` trait to execute commands, as @sminez indicated suggested, I added the `new_with_runner` method, which accepts both `&Paths` and `&impl CommandRunner`. The `new` method calls the `new_with_runner` method, using the `RealCommandRunner`.

And here is the dilemma:
- We make the `new_with_runner` method private or `pub(crate)`, which seems to make sense. However, as I understand (but not sure), it'll make mocking the `CommandRunner` in the integration tests impossible, since we won't be able to call the `pub(crate)` method from the `tests/` folder outside the source code (but I may be wrong)
- We make the `new_with_runner` method `pub`, but it'll require making the `CommandRunner` `pub` as well.

client/hwlib/src/collectors/os_info.rs
```diff
@@ -68,47 +81,90 @@ impl TryFrom<&Path> for KernelPackage {
...
+
+        // Mock the lsmod command for kernel package
+        mock_runner
+            .expect_run_command()
+            .withf(|cmd, args| cmd == "lsmod" && args.is_empty())
+            .returning(|_, _| Ok("Module Size Used\nsnd 61440 1\n".to_string()));
+
+        let result = OS::try_from((
+            get_test_filepath("version").as_path(),
+            &mock_runner as &dyn CommandRunner,
```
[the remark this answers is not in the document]
nadzyah: was removed!

client/hwlib/src/collectors/os_info.rs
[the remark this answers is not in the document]
TheSignPainter98: Ahhh I see, thanks for the detailed explanation! The only nice thing I can think of would be to move the integration tests into a `mod tests` in `lib.rs` as it's impossible to test without this internal detail. This is unfortunate, I agree that exposing this is awkward, but I don't think it's _too_ bad either way. For pragmatism, the existing visibility seems okay to me

sminez:
> @sminez Do we need to store a `Vec` of arguments in the hashmap? I think if we store a slice (`&'a [&'a str]`), the lifetimes still work out okay and this would remove the need for the `.map` call in `new`

If that works out fine lifetime wise (and I think you're correct that it should) then go for it 🙂 I was trying to avoid an overly line-noisy type signature for readability purposes but that's just personal preference on my part.

> Just two nitpicks with the above suggestion:
> - I think the generic lifetime parameter has quite a distinct purpose here, but the name `'a` doesn't really relate to any existing name to help convey it, how about renaming it to `'args`? (I'm a fan of descriptive lifetime names where it makes sense)

Descriptive lifetime names for public APIs I'm a fan of but in this case having `'args` along with your suggestion of using a slice instead of a vec would result it `(&'args str, &'args[&'args str])` which is pretty nasty readability wise(!)

> - I think instead of `MockRunner`, calling this `MockCommandRunner` would more clearly relate this back to `CommandRunner`

You could do but this is being defined and used immediately afterwards in the test rather than needing to be understood elsewhere in the codebase. So for similar arguments to the above I'd say keep the name short and sweet but again it comes down to preference here I think 🙂

TheSignPainter98:
@sminez I see what you mean, with both of my suggestions, that would be one very long line! I think the new type alias `CommandKey<'_>` helps alleviate this, but I think it's fine either way

For the naming of `MockRunner`, whilst it's true that it's being used immediately below, I don't know whether this will be used in other modules in future, but if it will be, I think matching the name would be beneficial. I see no clear cut answer here, I think both approaches have merit

sminez:
Yeah like I said this one very much feels like a personal preference thing. As things evolve over time there will probably be one way that looks better that the other but for now @nadzyah I'd say go with whatever you prefer 🙂

client/hwlib/src/collectors/os_info.rs
```diff
@@ -18,21 +18,35 @@
...
 use std::{fs::read_to_string, path::Path, process::Command};
 
 use crate::models::software::{KernelPackage, OS};
 
-impl TryFrom<&Path> for OS {
+pub trait CommandRunner {
+    fn run_command(&self, cmd: &str, args: &[&str]) -> Result<String>;
+}
+
+pub(crate) struct RealCommandRunner;
```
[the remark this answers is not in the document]
nadzyah: Updated in da1e20e

sminez (approved):

client/hwlib/src/collectors/os_info.rs
nadzyah: @TheSignPainter98 I've noticed when I use `vec`, I can specify and empty array, like `"lsmod", vec![]` and it works
When I replace the implementation with `type CommandKey<'args> = (&'args str, &'args [&'args str]);`, I get the following error:

```rust
("lsmod", &[]), Ok("Module Size Used\nsnd 61440 1\n")),
    |                        ^^^ expected an array with a fixed size of 1 element, found one with 0 elements
```

I don't feel like specifying `"lsmod", &[""]` is a good solution. What do you think?

client/hwlib/src/collectors/os_info.rs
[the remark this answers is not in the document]
nadzyah: I guess it's better to place the tests in `lib.rs` instead of exposing this trait. If we have issues with tests in future, we can make it `pub`

client/hwlib/src/collectors/os_info.rs
TheSignPainter98: Erm, that's very strange... is there a reference-to-array type somewhere (`&[T; _]` as opposed to a slice `&[T]`) which is forcing this? I think using a vec is okay here, as although it may allocate more, these are just tests, so it's okay 

This actually just gave me an idea (NB: optional, the existing solution is fine), I think it's possible to simplify the tests a little by making the keys to the map be just `Vec<&str>`/`&[&str]` rather than `(&str, Vec<&str>)`/`(&str, &[&str])` as although the latter makes it impossible to write a test without a command, if someone's done that, the test is broken anyway, so it's okay to panic. I think this would make specifying the input to the mock visually cleaner

client/hwlib/src/collectors/os_info.rs
```diff
@@ -18,21 +18,36 @@
...
 
-    fn try_from(proc_version_filepath: &Path) -> Result<Self> {
-        let codename = get_codename()?;
-        let distributor = get_distributor()?;
-        let version = get_version()?;
-        let kernel = KernelPackage::try_from(proc_version_filepath)?;
+impl OS {
+    pub(crate) fn try_new<C>(proc_version_filepath: &Path, runner: &C) -> Result<Self>
+    where
+        C: CommandRunner,
```
TheSignPainter98, line 45: I see a little inconsistency around parameters which implement `CommandRunner` are taken, specifically, we have:
- `fn some_fn<C>(runner: &C) where C: CommandRunner` and
- `fn some_fn(runner: &impl CommandRunner)`

I think we should try to have just one, I think `&impl CommandRunner` is nicer as it doesn't imply that an explicit type parameter is a requirement to define this functions type or implement it. (Also it doesn't require us to specify `C` in three places :P)

client/hwlib/src/collectors/os_info.rs
```diff
@@ -68,47 +83,104 @@ impl TryFrom<&Path> for KernelPackage {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::helpers::test_utils::get_test_filepath;
+    use std::collections::HashMap;
+
+    // Type aliases for better readability
+    type CommandKey<'args> = (&'args str, Vec<&'args str>);
```
TheSignPainter98, line 127: I think putting 'key' in the name of this type feels slightly off as it leaks an implementation detail---the current API specifically hides the use of the map behind the scenes. I think instead, we should focus on what this represents rather than how it's used; as we're mocking the `SystemCommandRunner`, how about calling this type `SystemCommand`?

client/hwlib/src/collectors/os_info.rs
```diff
@@ -68,47 +83,104 @@ impl TryFrom<&Path> for KernelPackage {
...
+    type CommandResult<'args> = Result<&'args str>;
```
TheSignPainter98, line 128: I think this type alias can be safely dropped as this obscures the true value of this type (I think putting the `Result<&'args str>` inline where required would be best as the name here doesn't provide additional clarity to me)

client/hwlib/src/collectors/os_info.rs
```diff
@@ -68,47 +83,104 @@ impl TryFrom<&Path> for KernelPackage {
...
+
+    struct MockCommandRunner<'args> {
+        calls: HashMap<CommandKey<'args>, CommandResult<'args>>,
```
TheSignPainter98, line 131: Hmmm, this `CommandResult<'args>` looks a little off in retrospect---I don't think the name `'args` represents the meaning of this lifetime too well here. Perhaps `'cmd` or just a one letter name would be better?

TheSignPainter98 (changes requested):
The logic is looking spotless, just a couple of notes getting the fine details nailed down, then I think this is ready

TheSignPainter98 (approved):
This is looking great Nadzeya, thanks for the changes!

PR #223, opened 2024-10-24
Implement component tests for the hwlib

nadzyah:
This PR adds the tests for checking the request body constructed by `CertificationStatusRequest` struct methods for case when SMBIOS data is available.

These tests are implemented in the `models::request_validators` module since we need to use mocking functionality that has `pub(create)` visibility. However, the tests check the whole request body and don't mock any internal functions or implementation, that's why these tests are component ones.

Additionally, the `MockCommandRunner` implementation was moved to `helpers::test_utils` to share this implementation between tests in different modules.


client/hwlib/src/helpers.rs
```diff
@@ -28,17 +28,52 @@ pub(crate) fn append_to_pathbuf(p: PathBuf, s: &str) -> PathBuf {
...
+
+    pub(crate) struct MockCommandRunner<'args> {
+        calls: HashMap<SystemCommand<'args>, Result<&'args str>>,
+    }
+
+    impl<'args> MockCommandRunner<'args> {
+        pub(crate) fn new(calls: Vec<(SystemCommand<'args>, Result<&'args str>)>) -> Self {
+            let calls = calls
+                .into_iter()
+                .map(|((cmd, args), res)| ((cmd, args.to_vec()), res))
```
sminez, line 50: You don't need this map here as `args` is already a Vec and if you remove `.to_vec()` this is just identity :slightly_smiling_face:

client/hwlib/src/helpers.rs
```diff
@@ -56,4 +91,13 @@ pub(crate) mod test_utils {
         }
         bail!("ran out of places to find Cargo.lock")
     }
+
+    pub(crate) fn get_expected_json(filepath: &Path, vars: &[(&str, &str)]) -> String {
+        let mut content = read_to_string(filepath).unwrap();
+        for (k, v) in vars {
+            content = content.replace(&format!("${k}"), v);
```
sminez, line 98: I should probably do the same in the prompting-client repo, but its probably worthwhile adding a comment here explaining what the expected template format being replaced here looks like (`$FOO` and `${FOO}`) given that the nested curly braces.

client/hwlib/src/models/request_validators.rs
```diff
@@ -173,3 +173,111 @@ impl CertificationStatusRequest {
...
+            "[{}]",
+            kernel_modules
+                .iter()
+                .map(|&module| format!("\"{}\"", module))
+                .fold(String::new(), |acc, x| if acc.is_empty() {
+                    x
+                } else {
+                    acc + ", " + &x
+                })
+        );
```
sminez, line 263: the `fold` here is essentially a `join` on a vec so you could make things a bit more readable by doing something like this:
```rust
        let quoted_kernel_modules: Vec<_> = kernel_modules
            .iter()
            .map(|&module| format!("\"{module}\""))
            .collect();

        let kernel_modules_str = format!("[{}]", quoted_kernel_modules.join(", "));
```

sminez (dismissed):
This is looking really good :muscle:

I've got a couple of comments around things that could be tidied up to make things more maintainable when you come back to these tests in the future but from our chat earlier today I think you have some follow on PRs planned so I'd suggest addressing these in the next PR rather than this one.

sminez (dismissed):
Very nice :ok_hand:

:shipit:

TheSignPainter98 (approved):
This is looking excellent, thanks both!

(Just one trivial :P)

PR #231, opened 2024-11-01
Update methods for constructing request to be arch specific

nadzyah:
This is a small PR that updates methods in `CertificationStatusRequest`
Instead of having `from_smbios_data` for the amd64 architecture and `from_defaults` for others, now we use conditional compilation for the `from` method.

### DIFF
--- a/client/hwlib/src/models/request_validators.rs
+++ b/client/hwlib/src/models/request_validators.rs
@@ -20,15 +20,10 @@
 
 use anyhow::{anyhow, Result};
 use serde::{Deserialize, Serialize};
-use smbioslib::{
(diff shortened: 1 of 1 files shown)


sminez (approved):
Nice and clean :ok_hand: 

The only thing I think you'll want to keep and eye on / think about going forward is how this affects your ability to test the different moving parts here. So long as you have test coverage that isn't also guarded by `#[cfg(target_arch = ...)]` you should be ok, but if the tests are also conditionally run based on the target architecture you're going to end up with CI not quite doing what you expect in a few cases I suspect? That or you'll need to make sure that you are explicitly running on different architectures :slightly_smiling_face:

PR #235, opened 2024-11-06
Improve server logic

nadzyah:
This PR updates the server logic after testing it on certified machines using scripts in #234. The patches contain the following fixes:

* Use a list of BIOSes instead of `first()`. Since we match only bios vendor and version, there might be multiple BIOSes with different revisions but the same version. One of them may belong to the matching machine. So, now we use all matching BIOSes instead of getting the first one. Additionally, we don't match other fields like `revision` since firmware may receive revision updates pretty frequently, but it still allows us to consider hardware as potentially certified since we're still working with the same BIOS version
* Checking the board version turned out to be a strict validation. After the discussion with the team, it was decided to check only the board vendor and model
* It turned out that the `lsb_release` command used by the client to collect the OS version returns the string in format `22.04` without the `LTS` suffix. However, C3 stores it with the `LTS` suffix. The logic was added to remove this suffix while importing the data from C3
* The hardware mismatches are logged now

Also, the tests were revised/added to cover the implemented logic

### DIFF
--- a/server/hwapi/endpoints/certification/certification.py
+++ b/server/hwapi/endpoints/certification/certification.py
@@ -64,11 +64,10 @@ def check_certification(
 
     # Match against board and bios
     try:
-        board, bios_list = logic.find_main_hardware_components(
-            db, system_info.board, system_info.bios
-        )
+        board = logic.find_board(db, system_info.board)
+        bioses = logic.find_bioses(db, system_info.bios) if system_info.bios else []
         related_machine = logic.find_certified_machine(
-            db, system_info.architecture, board, bios_list
+            db, system_info.architecture, board, bioses
         )
     except ValueError:
         logging.error(
--- a/server/hwapi/endpoints/certification/certification.py
+++ b/server/hwapi/endpoints/certification/certification.py
@@ -18,7 +18,7 @@
 #        Nadzeya Hutsko <nadzeya.hutsko@canonical.com>
 """The endpoints for working with certification status"""
 
-
+import logging
 from fastapi import APIRouter, Depends
 from sqlalchemy.orm import Session
 
@@ -61,6 +61,7 @@ def check_certification(
     vendor = repository.get_vendor_by_name(db, system_info.vendor)
     if not vendor:
         return NotCertifiedResponse()
+
     # Match against board and bios
     try:
         board, bios_list = logic.find_main_hardware_components(
@@ -70,16 +71,30 @@ def check_certification(
             db, system_info.architecture, board, bios_list
         )
     except ValueError:
+        logging.error(
+            (
+                "Hardware cannot be found. Machine vendor: %s, model: %s"
+                ", board model: %s, board version: %s, bios version: %s"
+            ),
+            system_info.vendor,
+            system_info.model,
+            system_info.board.product_name,
+            system_info.board.version,
+            system_info.bios.version if system_info.bios else None,
+        )
         return NotCertifiedResponse()
+
     bios = repository.get_machine_bios(db, related_machine.id)
     related_releases, kernels = repository.get_releases_and_kernels_for_machine(
         db, related_machine.id
     )
+
     # Match against CPU codename
     if not logic.check_cpu_compatibility(db, related_machine, system_info.processor):
         return response_builders.build_related_certified_response(
             db, related_machine, board, bios, related_releases, kernels
         )
+
     # Check OS release
     release_from_request = repository.get_release_object(
         db, system_info.os.version, system_info.os.codename
--- a/server/hwapi/endpoints/certification/certification.py
+++ b/server/hwapi/endpoints/certification/certification.py
@@ -63,14 +63,15 @@ def check_certification(
         return NotCertifiedResponse()
     # Match against board and bios
     try:
-        board, bios = logic.find_main_hardware_components(
+        board, bios_list = logic.find_main_hardware_components(
             db, system_info.board, system_info.bios
         )
         related_machine = logic.find_certified_machine(
-            db, system_info.architecture, board, bios
+            db, system_info.architecture, board, bios_list
         )
     except ValueError:
         return NotCertifiedResponse()
+    bios = repository.get_machine_bios(db, related_machine.id)
     related_releases, kernels = repository.get_releases_and_kernels_for_machine(
         db, related_machine.id
     )
--- a/server/hwapi/endpoints/certification/logic.py
+++ b/server/hwapi/endpoints/certification/logic.py
@@ -18,7 +18,6 @@
 #        Nadzeya Hutsko <nadzeya.hutsko@canonical.com>
 """The algorithms for determining certification status"""
 
-from typing import Sequence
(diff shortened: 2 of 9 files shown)


omar-selo:
Reading the PR description, I wonder if it's even necessary to match BIOS when checking if a hardware is certified. As in does the BIOS used affect whether Ubuntu would run fine or not?

server/tests/endpoints/certification/test_certification.py
```diff
@@ -94,7 +96,66 @@ def test_hardware_mismatch(generator: DataGenerator, test_client: TestClient):
...
+    """
+    vendor = generator.gen_vendor(name="Known Vendor")
+    board = generator.gen_device(
+        vendor,
+        identifier="dmi:0001",
+        name="FW00Q",
+        version="0001A",
+        category=DeviceCategory.BOARD,
+        bus=BusType.dmi,
+    )
```
omar-selo, line 136: It would be great if the generator had more specific device creation methods. For instance `gen_board`. This way we could prefill the category and we could also add some defaults that would generate a dummy board for us without needing to repeat it in every test. Doing that, this code would just look like this:
```python
board = generator.gen_device(vendor)
```

server/tests/endpoints/certification/test_certification.py
```diff
@@ -94,7 +96,66 @@ def test_hardware_mismatch(generator: DataGenerator, test_client: TestClient):
...
+    generator.gen_bios(vendor=vendor, version="1.0.12")
+
+    response = test_client.post(
+        "/v1/certification/status",
+        json={
```
omar-selo, line 141: This post is repeated across tests perhaps it's an opportunity for following the DRY principle. We could create a class with a helper method that returns some default json request prefilled with some dummy data. This function can take as input some optional parameters like vendor name and output a customized json in such situations. If we do that, then we avoid the need to repeat this structure in every test with dummy values. More importantly we decouple the structure of the request from the tests, such that if the structure changes then we only need to make a change in one place.

server/tests/endpoints/certification/test_certification.py
```diff
@@ -110,6 +171,8 @@ def test_hardware_mismatch(generator: DataGenerator, test_client: TestClient):
 
     assert response.status_code == 200
     assert response.json() == {"status": "Not Seen"}
+    assert "Hardware cannot be found" in caplog.text
+    assert "bios version: 1.0.15" in caplog.text
```
omar-selo, line 175: I see that the logs are being used to test that the reason hardware is "Not Seen" is what's expected of this test. It's great that we're testing this. But I suggest that instead of depending on logs, we test it directly. This can be done by refactoring the logic that checks if hardware is certified in it's own function. This function would be used by the controller. It can return not only the status but also the reason. Then we can write tests for that function to make sure the reason matches what we expected.

server/tests/endpoints/certification/test_certification.py
```diff
@@ -504,30 +567,7 @@ def test_bios_is_none(generator: DataGenerator, test_client: TestClient):
...
-                "codename": release.codename,
-                "kernel": {
-                    "name": report.kernel.name,
-                    "version": report.kernel.version,
-                    "signature": report.kernel.signature,
-                    "loaded_modules": [],
-                },
-            }
-        ],
-    }
```
omar-selo, line 530: Is there a particular reason for modifying the test to check only for the field "status" being "Certified"? It seems like we're weakening the test this way. For instance if some code change causes other json changes then this test will no longer catch it. I'd propose we always make sure the whole output is correct. We can avoid making this check every time by using the same helper class we defined earlier that generated the status request. We can add three methods to this class that takes the output returned and checks if it's the same as the expected output (that it knows because it generated the status request initially). These methods would be something like check certified, check partial certified and check not certified. The reason for having three methods is to avoid adding logic to decide whether it's certified or not.

server/hwapi/endpoints/certification/logic.py
```diff
@@ -31,30 +31,29 @@
 
 def find_main_hardware_components(
     db: Session, board_data: BoardValidator, bios_data: BiosValidator | None
-) -> tuple[models.Device, models.Bios | None]:
+) -> tuple[models.Device, Sequence[models.Bios]]:
```
omar-selo, line 34: - why not separate this into two functions, one to find the board and another for bios?
- it's encouraged to be generic with what you accept and specific with what you return. In other words while it makes sense to take as input an abstract type like a Sequence, we should return a concrete type like a list. This is because the user of the function should know exactly what kind of object they can expect returned.

server/hwapi/endpoints/certification/logic.py
```diff
@@ -18,7 +18,7 @@
 #        Nadzeya Hutsko <nadzeya.hutsko@canonical.com>
 """The algorithms for determining certification status"""
 
-
+from typing import Sequence
```
omar-selo, line 21: starting from Python 3.9 typing.T is deprecated for collections.abc.T https://stackoverflow.com/questions/75285711/difference-between-collections-abc-sequence-and-typing-sequence

omar-selo (changes requested):
Thanks for the changes. Just had a few comments. Main ones are around adding a testing helper class and avoiding logging for testing

server/hwapi/endpoints/certification/certification.py
```diff
@@ -61,24 +61,40 @@ def check_certification(
...
+            (
+                "Hardware cannot be found. Machine vendor: %s, model: %s"
+                ", board model: %s, board version: %s, bios version: %s"
+            ),
+            system_info.vendor,
+            system_info.model,
+            system_info.board.product_name,
+            system_info.board.version,
+            system_info.bios.version if system_info.bios else None,
+        )
```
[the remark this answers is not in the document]
nadzyah: Matias specifically asked for logging hardware mismatch data since we'll probably have to adjust the logic according to, for instance, how many certified machines run a newer bios version, different from which we used for issuing a certificate

server/tests/endpoints/certification/test_certification.py
nadzyah: Matias asked to check that if there is a hardware mismatch, we log this information

server/hwapi/endpoints/certification/certification.py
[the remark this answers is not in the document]
omar-selo: This sounds like we need to store this data to be able to query it rather than just log it. I'm not sure if simply logging will help us get such information. @mz2 wdyt?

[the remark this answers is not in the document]
mz2: Fair point, it probably is better that way. I was imagining deriving metrics from logs, but explicitly storing the full data is probably a better idea.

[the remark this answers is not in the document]
nadzyah: After discussing it with @omar-selo, we've decided to add this change in another PR and create a table for storing hardware mismatches

omar-selo (approved):
Looks good! and thanks addressing my comments!

PR #238, opened 2024-11-13
Add integration tests

nadzyah:
This PR adds simple integration tests that check how the server and client communicate.
It also provides a GitHub CI pipeline for running them on PRs or pushes to main


tests/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    pkgconf \
+    libssl-dev \
+    lsb-release \
+    kmod \
+    python3 \
+    python3-pip \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Rust
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
```
plars, line 16: It might be better to avoid the pipe to sh install method since we can also just `apt install rustup`. You can also omit the next line to set the path and do `rustup default stable` instead. Both methods seem to result in getting 1.82.0 for me.

plars (commented):
I haven't completely tried to run these yet, just a drive-by comment :)

tests/Dockerfile
nadzyah: This method is recommended by Rust Team: https://www.rust-lang.org/tools/install
Do you think we should use apt instead?

plars: It's an 800+ line shell script that you pull from an external source, vs a package that is reviewed and signed off by a small group of individuals that have upload access to the Ubuntu archives.

It's "recommended" because it's the easiest thing that they can do to reach the widest audience regardless of what OS/distro they use, but in general, it's probably better to find another option if there is one. The deb package here gives you the same version of rust in the end too.

tests/src/integration_test.rs
```diff
@@ -0,0 +1,110 @@
...
+    let cert_request = CertificationStatusRequest::new(get_test_device_paths(dir_path))?;
+    let response = send_certification_status_request(api_url, &cert_request).await?;
+
+    match response {
+        CertificationStatusResponse::Certified {
+            architecture,
+            bios: _,
+            board: _,
+            chassis: _,
+            available_releases,
```
sminez, line 67: rather than underscoring the variables you don't need you can use `..` to discard them from the match:
```rust
CertificationStatusResponse::Certified {
    architecture,
    available_releases,
    ..
},
```

tests/Dockerfile
sminez: Alternatively since this is just for running the tests themselves, there is an [official rust image](https://hub.docker.com/_/rust) that you could use as a base and then all you should need to do is copy over your tests set your entry point?

nadzyah: I've tried it but it's debian-based and we need ubuntu-based environment

tests/src/integration_test.rs
```diff
@@ -0,0 +1,110 @@
...
+    "thinkstation_p620"
+)]
+#[tokio::test]
+async fn test_certification_request(dir_path: &str) -> Result<()> {
+    let api_url = std::env::var("API_URL")
+        .unwrap_or_else(|_| String::from("http://hwapi-server-integration:8080"));
+    let cert_request = CertificationStatusRequest::new(get_test_device_paths(dir_path))?;
+    let response = send_certification_status_request(api_url, &cert_request).await?;
+
+    match response {
```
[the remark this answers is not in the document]
nadzyah: I want to add this logic as the next step, which is implementing snapshot tests. Here, I just set up the environment for them and added basic tests. With the snapshot tests, I'm planning to populate the test DB with sample data, and then, of course, we'll be able to match a specific response body

tests/Dockerfile
nadzyah: I've updated the Dockerfile to use `apt install rustup`

sminez (approved):
Looking good on the Rust side now :slightly_smiling_face:

PR #247, opened 2024-11-22
Make integration tests to be snapshot tests

nadzyah:
This PR updates the project's integration tests and makes them snapshot tests. Information about the tests can be found in the added [README.md](https://github.com/canonical/hardware-api/blob/snapshot-tests/integration-tests/README.md) file.


server/hwapi/data_models/repository.py
```diff
@@ -84,7 +84,9 @@ def get_board(db: Session, vendor_name: str, product_name: str) -> models.Device
         .join(models.Vendor)
         .where(
             and_(
-                models.Vendor.name.ilike(_clean_vendor_name(vendor_name)),
+                models.Vendor.name.ilike(
+                    f"%{_clean_vendor_name(vendor_name).lower()}%"
+                ),
```
omar-selo, line 89: why not add `.lower` to `_clean_vendor_name` function?

omar-selo:
I've been thinking about this over the weekend and I think it would make sense if these tests which run part as part of hardware API CI only test hardware API and not C3 in any way. If we want an integration test between C3 and hardware API that should be triggered whenever a change occurs in this repo or C3's. But in this case this integration test is just for hardware API. So perhaps we shouldn't fetch anything from C3. Instead it would make sense to go with the option you proposed to me on MM in which we store the expected partial C3 response. This way we'd test hardware API all the way starting from retrieving data from C3 (albeit mocked). wdyt?

server/scripts/import_test_data_from_c3.py
omar-selo, line 1: what do you think of moving `import_test_data_from_c3.py` and the related json files from `c3_test_data` into `integration-tests` directory?

omar-selo (changes requested):

server/scripts/import_test_data_from_c3.py
nadzyah: it would be challenging since the docker container that we set up for the server works in the `server/` directory context

omar-selo: Can you rename the file `import_test_data_from_c3`? it's a bit misleading

nadzyah: what name do you prefer?

PR #253, opened 2024-12-03
Pack client deb hwlib hwctl

nadzyah:
This PR closes [C3-913](https://warthogs.atlassian.net/browse/C3-913).
With these changes, the source client package now provides two binaries: `librust-hwlib-dev` and `hwctl`. To make such packaging possible, the `debian/` folder and workspace `Cargo.*` files were moved to the `client/` folder.

Key changes:

* The `debian/rules` file was updated with an overridden dh_auto_install section. It was done `dh-cargo` treats the whole workspace as a single package, either library or binary package. For more information, check https://salsa.debian.org/rust-team/dh-cargo/-/blob/5cc7f7b8/cargo.pm#L168-195
* The `debian/generate-checksum.py` script was updated to generate the `cargo-checksum.json` file only for the files in the `hwlib` dir, since it's needed only for the library packages
* `hwctl.1` man page was added since it's required for CLI tools

[C3-913]: https://warthogs.atlassian.net/browse/C3-913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zyga (commented):
Some feedback on the man page.

Super nice for you to write it manually. I really applaud that!

.github/workflows/test_client.yaml
```diff
@@ -20,6 +20,9 @@ concurrency:
 jobs:
   lint-and-test:
     runs-on: [self-hosted, linux, large, jammy, x64]
+    defaults:
+      run:
+        working-directory: client
```
zyga, line 25: TIL

client/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[workspace]
+members = [
+    "hwctl",
+    "hwlib"
+]
+resolver = "2"
+exclude = ["rust-vendor",]
+
+[workspace.metadata.vendor-filter]
+platforms = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu"]
```
zyga, line 10: Do we certify any other architectures where this this could be a problem?

client/debian/control
```diff
@@ -4,24 +4,53 @@ Priority: optional
...
+Vcs-Browser: https://github.com/canonical/hardware-api/tree/main/client
 Homepage: https://github.com/canonical/hardware-api
 X-Cargo-Crate: hwlib
 Rules-Requires-Root: no
 
 Package: librust-hwlib-dev
-Architecture: any
-Multi-Arch: same
-Depends: libssl-dev, pkgconf, ${misc:Depends}, ${shlibs:Depends}
+Architecture: amd64
```
zyga, line 23: This looks wrong. Why would you constrain the architecture?

client/debian/control
```diff
@@ -4,24 +4,53 @@ Priority: optional
...
+ communicates with the certification database to check:
+  * If the exact or similar device model has been certified
+  * Which Ubuntu releases the device is certified for
+  * Current certification status
+ .
+ This package provides the development files needed to use hwlib in Rust
+ applications.
+
+Package: hwctl
+Architecture: amd64
```
zyga, line 42: Ditto

client/debian/control
```diff
@@ -4,24 +4,53 @@ Priority: optional
...
+Multi-Arch: foreign
+Depends: ${misc:Depends}, ${shlibs:Depends}
+Description: Command-line tool for checking Ubuntu hardware certification status - Rust source code
```
zyga, line 45: Please check if this field should be capitalize. I don't recall now.

client/debian/rules
```diff
@@ -0,0 +1,59 @@
...
+# Defines DEB_*_RUST_TYPE triples
+include /usr/share/rustc/architecture.mk
+
+export CARGO_HOME = $(CURDIR)/debian/
+export CARGO_VENDOR_DIR = $(CURDIR)/rust-vendor/
+# Enable debug symbols to avoid debug-file-with-no-debug-symbols warning
+export CARGO_PROFILE_RELEASE_DEBUG=1
+
+CARGO_TARGET_DIR = $(CURDIR)/target
+PARALLEL := $(shell nproc)
```
zyga, line 12: I believe there is an environment variable for exactly this purpose, but I don't recall its name.

client/debian/rules
```diff
@@ -0,0 +1,59 @@
...
+		/usr/bin/cargo -Zavoid-dev-deps install \
+		--verbose \
+		--verbose \
+		-j$(PARALLEL) \
+		--target $(DEB_HOST_RUST_TYPE) \
+		--config 'profile.release.lto = "thin"' \
+		--no-track \
+		--path hwctl \
+		--root $(CURDIR)/debian/hwctl/usr
+	install -D -m 644 Cargo.lock $(PKG_HWCTL)/usr/share/doc/hwctl/Cargo.lock
```
zyga, line 53: What is the point of installing the lock file?

client/hwctl.1
```diff
@@ -0,0 +1,48 @@
...
+The
+.Nm
+utility does not accept any command-line options.
+.Sh ENVIRONMENT
+.Bl -tag -width "HW_API_URL" -compact
+.It Ev HW_API_URL
+Specifies the hardware-api server URL that
+.Nm
+uses for certification queries. The default value is
+.Pa https://hw.ubuntu.com .
```
zyga, line 36:
```suggestion
. Lk https://hw.ubuntu.com .
```

zyga (commented):
I did a full pass and many things look right, but I left a few questions. Please have a look.

client/Cargo.toml
nadzyah: currently, the library supports amd64 architecture and has limited support for arm64. that's why we pass both these architectures here, but pack only for amd64 since this is the only architecture the client is ready to be used on.

client/debian/control
nadzyah: This is the only architecture we fully support now, since we primarily use SMBios for collecting hardware data, which works only on amd64 architectures
I've checked that `dmidecode` tool, which uses SMBios as well, also specifies architecture explicitly in its `debian/control` file

client/debian/control
nadzyah: Seems that all Rust libraries have it capitalised
image

client/debian/rules
nadzyah: MIR requires it:

```
TODO-C: - This package uses vendored rust code tracked in Cargo.lock as shipped,
TODO-C:   in the package (at /usr/share/doc/<pkgname>/Cargo.lock - might be
TODO-C:   compressed)
```

client/hwctl.1
nadzyah: But it's already shown as a link on the man page
image

zyga: Yes but .Pa says "this is a path". The link is rendered by your terminal emulator by matching text.

client/debian/rules
zyga: Thanks, I think you should include an explanation in the code as well.

client/debian/control
zyga: Thanks for checking.

client/debian/control
zyga: Ack.

client/Cargo.toml
zyga: Ack

zyga (approved):
LGTM but please add a comment about lock file and see my comment about why .Pa is unsuitable for URLs.

Nice work!

client/debian/rules
nadzyah: added

PR #260, opened 2024-12-06
Add user-friendly error-handling to hwctl

nadzyah:
Currently, when you run `hwctl` binary without root permissions or on a system with no SMBIOS data available, you get the following output:
```
failed to load SMBIOS data: Permission denied (os error 13).
thread 'main' panicked at hwlib/src/models/request_validators.rs:99:85:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This happens because there is no proper error-handling functionality in the `CertificationStatusRequest::from` method that loads data from SMBIOS table.

This PR adds the proper error-handling functionality, so the output looks the following way:

```
failed to collect system data: Permission denied (os error 13)
ensure you're root and running the program on a machine with SMBIOS available
```


client/hwctl/src/main.rs
```diff
@@ -41,8 +50,8 @@ async fn main() -> Result<()> {
             exit(0);
         }
         Err(e) => {
-            eprintln!("{}", e);
-            exit(1);
+            eprintln!("failed to send certification status request: {}", e);
+            exit(2);
```
TheSignPainter98, line 54: Are the different exit codes significant? I think it may be cleaner to use `ExitCode::SUCCESS` and `ExitCode::FAILURE` otherwise (from `std::process::exit`).

Also, just to highlight that `exit` isn't defined in this file, I think it would be nice to prefix it with part of the path it's defined in. Not sure how to word it exactly, essentially this:
```rust
use std::process;
// ...
process::exit(...);
```
It may be better to do this in another PR though as I suspect there may be places outside of the current changes which would need to be updated. For this PR, it's probably okay just to follow the existing convention

client/hwctl/src/main.rs
```diff
@@ -27,7 +27,16 @@ use hwlib::{
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    let cert_status_request = CertificationStatusRequest::new(Paths::default())?;
+    let cert_status_request = match CertificationStatusRequest::new(Paths::default()) {
+        Ok(request_data) => request_data,
+        Err(e) => {
+            eprintln!("failed to collect system data: {}", e);
```
TheSignPainter98, line 33: Just to follow [the convention](https://github.com/canonical/rust-best-practices/blob/main/best-practices.md#error-messages), I think this error message should start with the word 'cannot' rather than 'failed to.' Also, I think it would be nice to inline that format parameter, as in `"... {e}"`

client/hwctl/src/main.rs
```diff
@@ -27,7 +27,16 @@ use hwlib::{
...
+            eprintln!(
+                "ensure you're root and running the program on a machine with SMBIOS available"
```
TheSignPainter98, line 35: I think this advice isn't very extensible, as the problem the user encounters may not relate to the user being the root or the availability of SMBOIS data, in which case this is misleading. I think in this case, it's okay to rely on the user being smart enough to see something like `cannot access file foo: permission denied` and understand that they need to run as the root. In the case of missing SMBIOS data, I don't think there's anything the user can do apart from buy a new computer? :P 

Minor note: I think these two messages use two names for the same thing---are 'system data' and 'SMBIOS' equivalent in the eyes of hwctl? If so, I think it'd be better to use only one name to reduce cognitive overhead on the user

Just for future reference, in these kind of advice / error messages, I think it's good practice to avoid the second person and instead to focus on the program itself as addressing the user directly in this way feels informal. In this case, we could use:
```
ensure that hwctl is running as root
```

client/hwctl/src/main.rs
```diff
@@ -27,7 +27,16 @@ use hwlib::{
 
 #[tokio::main]
 async fn main() -> Result<()> {
```
TheSignPainter98, line 29: This may be beyond the scope of this PR, but there are currently two ways for errors to be reported, either explicitly with the new `eprintln!` calls, or implicitly by returning `Err` from `main`. In the latter case, the error message will look different (and I think often uglier?), so it's probably best to write another function which catches all errors and outputs them in a unified way:
```rust
async fn main() -> ExitCode {
    let err = match run().err() {
        Some(err) => err,
        None => return ExitCode::SUCCESS,
    };
    eprintln!("{e}");
    ExitCode::FAILURE
}

fn run() -> Result<()> {
    // Everything currently in main
}

File: client/hwlib/src/models/request_validators.rs [hunk 5]
```diff
@@ -96,7 +96,12 @@ impl CertificationStatusRequest {
             proc_version_filepath,
             ..
         } = paths;
-        let data = load_smbios_data(&smbios_entry_filepath, &smbios_table_filepath).unwrap();
+        let data = match load_smbios_data(&smbios_entry_filepath, &smbios_table_filepath) {
+            Ok(smbios_data) => smbios_data,
+            Err(e) => {
+                return Err(e);
+            }
+        };
```
TheSignPainter98, line 104: I think this `match` may be equivalent to `?` :)

client/hwctl/src/main.rs
```diff
@@ -41,8 +50,8 @@ async fn main() -> Result<()> {
             exit(0);
         }
         Err(e) => {
-            eprintln!("{}", e);
-            exit(1);
+            eprintln!("failed to send certification status request: {}", e);
```
TheSignPainter98, line 53: I think we should `s/failed to/cannot/` here

client/hwlib/src/collectors/hardware_info.rs
```diff
@@ -31,14 +31,11 @@ use crate::{
...
-        Ok(data) => Some(data),
-        Err(e) => {
-            eprintln!("failed to load SMBIOS data: {}.", e);
-            None
-        }
-    }
+pub(crate) fn load_smbios_data(
+    entry_filepath: &Path,
+    table_filepath: &Path,
+) -> Result<SMBiosData, anyhow::Error> {
```
TheSignPainter98, line 37: I think the error type of this `Result` is equivalent to an `anyhow::Result`, so it's probably cleaner just to use that alias :)

TheSignPainter98 (changes requested):
This is looking very nice! Just a couple of tweaks

client/hwctl/src/main.rs
nadzyah: I've decided to keep it in one function since otherwise, I feel like the `run` function will do almost nothing
isn't the purpose of `main` to call the functions in the correct order, handling errors if they appear?

client/hwctl/src/main.rs
```diff
@@ -17,32 +17,43 @@
...
-    );
-    let url = std::env::var("HW_API_URL").unwrap_or_else(|_| String::from("https://hw.ubuntu.com"));
-    let response = send_certification_status_request(url, &cert_status_request).await;
-
-    match response {
-        Ok(response) => {
-            println!("\nResponse:\n{}", serde_json::to_string_pretty(&response)?);
-            exit(0);
-        }
+async fn main() -> process::ExitCode {
```
TheSignPainter98, line 29: Just for consistency, as `ExitCode` is a type and doesn't require any disambiguating, I think it's best to import it directly

client/hwctl/src/main.rs
```diff
@@ -17,32 +17,43 @@
...
+            eprintln!("ERROR: {e:?}");
+            process::ExitCode::FAILURE
         }
     }
 }
+
+async fn run() -> Result<()> {
+    let cert_status_request = CertificationStatusRequest::new(Paths::default()).context(
+        "cannot collect system data. \
+         Ensure that hwctl is running as root and on an SMBIOS-compatible system.",
```
TheSignPainter98, line 42: Just from a comment in my last review, I think we should remove the second part of the second sentence, and possibly the whole second sentence entirely

client/hwctl/src/main.rs
```diff
@@ -17,32 +17,43 @@
...
+    )?;
+
+    let request_json = serde_json::to_string_pretty(&cert_status_request)
+        .context("cannot serialize request to JSON")?;
+    println!("Request:\n{}", request_json);
+
+    let url = std::env::var("HW_API_URL").unwrap_or_else(|_| String::from("https://hw.ubuntu.com"));
```
TheSignPainter98, line 49: This isn't a massive issue, but that `String::from` requiring us to use a closure to prevent unnecessary allocations seems a little clunky. Instead, is it possible to handle an `Option<&str>` rather than an `Option<String>`? (I don't recall the type of the first parameter to `send_certification_status_request`)

```rust
let url = env::var("HW_API_URL").as_deref().unwrap_or("https://hw.ubuntu.com");
```
Also, I think it would be slightly nicer to use the `var` function as `env::var` rather than `std::env::var` (the `std` part doesn't seem too important)

client/hwlib/src/collectors/hardware_info.rs
```diff
@@ -31,14 +31,8 @@ use crate::{
...
-pub(crate) fn load_smbios_data(entry_filepath: &Path, table_filepath: &Path) -> Option<SMBiosData> {
-    match table_load_from_device(entry_filepath, table_filepath) {
-        Ok(data) => Some(data),
-        Err(e) => {
-            eprintln!("failed to load SMBIOS data: {}.", e);
-            None
-        }
-    }
+pub(crate) fn load_smbios_data(entry_filepath: &Path, table_filepath: &Path) -> Result<SMBiosData> {
+    table_load_from_device(entry_filepath, table_filepath)
```
TheSignPainter98, line 35: This function seems somewhat redundant now, would it be cleaner to move the content of `table_load_from_device` here? I'm not sure whether this is used elsewhere

TheSignPainter98 (changes requested):
This is looking really good, thanks for the changes! Just a few last bits and pieces then this is ready to go

client/hwctl/src/main.rs
nadzyah: If I use `&str` instead of `String`, I get this error
`creates a temporary value which is freed while still in use`

TheSignPainter98: Ahhh yes of course, the `Option<String> ` from `env::var` would require a binding, my mistake. That seems like too much, no worries, I think this code is probably cleanest as-is (this is a one-time cost anyway)

TheSignPainter98 (approved):
This is looking very nice, thanks for the changes!

PR #261, opened 2024-12-10
Automate versioning scheme for the client cargo and Debian packages

nadzyah:
The PR resolves C3-957. 
It adds the automation pipeline to ensure that the Debian changelog main version (before the `~` sign) and cargo version for both `hwlib` and `hwctl` packages are the same and match the latest `v*.*.*` version.

Since the server charm publishing workflow also creates tags, it was adjusted to use a tag_prefix to make the charm releases easily distinguishable from the client tags in the repo.


client/README.md
```diff
@@ -121,7 +124,46 @@ dch -i  # increment release number
...
+
+3. Create a Matching Git Tag
+
+Ensure the version is correctly tagged in Git. The tag should follow
+the format `vMAJOR.MINOR.PATCH` and must be annotated to the same
+commit as the changelog entry:
+
+```
+git tag -a v1.2.3 -m "Release version 1.2.3"
+git push origin v1.2.3
```
plars, line 150: Should this tag be against a specific revision? for instance, the one where the toml files and the debian changelog are updated?

plars (commented):
one small question

client/README.md
nadzyah: yes, we should tag the commit that changes versions in the changelog and toml files

client/README.md
```diff
@@ -121,7 +124,46 @@ dch -i  # increment release number
...
+```
+
+4. Update Cargo Versions
+
+If the new version includes changes to the MAJOR, MINOR, or PATCH
+version numbers, update the version fields in the following files:
```
boukeas, line 156: I am little confused. Is there any chance there would be a release that doesn't change at least one of these numbers?

boukeas (approved):
+1

I have just one question, something that might require a rephrasing or clarification in the documentation.

client/README.md
nadzyah: you can release an update to PPA with the `ppa` prefix, but if you do releases to Ubuntu main, you should increment the version number

client/README.md
nadzyah: I've updated the instructions to indication that we specify a commit hash with the `git tag` command

PR #269, opened 2025-01-07
Update dependencies

nadzyah:
This PR updates both server and client dependencies.


plars (approved):
really weird that it needed the README! wow, +1

PR #279, opened 2025-02-06
Update client dependencies and pack it for plucky

nadzyah:
This PR updates client dependencies and sets the latest target release to be plucky instead of oracular


renanrodrigo (approved):
I see this was built and tested in https://launchpad.net/~nhutsko/+archive/ubuntu/testppa
- rust-hwlib/0.9.0~ppa2
    + ✅ rust-hwlib on plucky for amd64   @ 11.02.25 14:59:58

LGTM

PR #287, opened 2025-02-17
In hwctl command, don't output the request information

nadzyah:
Since a user is interested only in the status information, we should avoid outputting the request information.

### DIFF
--- a/client/debian/changelog
+++ b/client/debian/changelog
@@ -1,3 +1,9 @@
+rust-hwlib (0.9.0~ppa3) plucky; urgency=medium
+
+  * Don't output the request information in hwctl.
+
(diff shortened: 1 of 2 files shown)


client/debian/changelog
```diff
@@ -1,3 +1,9 @@
+rust-hwlib (0.9.0~ppa3) UNRELEASED; urgency=medium
```
renanrodrigo, line 1:
```suggestion
rust-hwlib (0.9.0~ppa3) plucky; urgency=medium
```

renanrodrigo: As a rule of thumb, it's better if changelog/release related changes live in separate commits. Your call though (:

renanrodrigo (approved):

PR #297, opened 2025-04-24
Use full paths of binaries for OS info collection and fix dependency issues

nadzyah:
It was recommended by the MIR reviewers to use the full paths of binaries: https://bugs.launchpad.net/ubuntu/+bug/2072561/comments/10


client/hwlib/src/constants.rs
```diff
@@ -22,3 +22,7 @@ pub const CPU_MAX_FREQ_FILE_PATH: &str = "/sys/devices/system/cpu/cpu0/cpufreq/c
 pub const PROC_CPUINFO_FILE_PATH: &str = "/proc/cpuinfo";
 pub const PROC_DEVICE_TREE_DIR_PATH: &str = "/proc/device-tree/";
 pub const PROC_VERSION_FILE_PATH: &str = "/proc/version";
+
+pub const DPKG: &str = "/usr/bin/dpkg";
+pub const LSB_RELEASE: &str = "/usr/bin/lsb_release";
+pub const LSMOD: &str = "/usr/bin/lsmod";
```
TheSignPainter98, line 28: I think it's worth adding a suffix to clarify what these constants are---rather than being these programs themselves, they're actually commands/paths. Perhaps calling these like `LSMOD_CMD` or `LSMOD_PATH` would make the intention even clearer?

TheSignPainter98 (changes requested):
This is looking very nice, just one trivial

client/hwlib/src/constants.rs
nadzyah: We're not planning to use other versions of these commands other than the full path since this is a security recommendation.
I think it makes sense to keep them without any suffix showing that this is the only was to call, for example, `lsmod` command

TheSignPainter98: Fair enough, I think it's not worth blocking on this issue. Either way seems reasonable

TheSignPainter98 (approved):
The changes look good, in the interest of time, I'm pre-approving this, pending resolution (either way) of the one trivial noted above

PR #304, opened 2025-05-01
Improve autopkg tests to resolve all MIR TODOs

nadzyah:
This PR adds integration auto tests to resolve all the remaining MIR TODOs.

The autopkgtests TODO states the following:

```
#2 - The autopkgtests is just a re-execution of the unit tests, including
     re-compile, so not testing the actual, installed binaries. Adding a simple
     integration tests, using a mock HTTP server to simulate hw.ubuntu.com and
     make a test request using the real binary would be very valuable.
```

So, here are the main changes in this PR:
- Add the hwctl test. It mocks HTTP server that returns a static `Not Seen` status regardless of the request body
- Make sure that client doesn't fail if some hardware information is not available
- Require `lsb_release` and `kmod` when installing `hwctl`
- Update cargo depdendenies (required by the Security team)
- Add instructions on running autopkgtests for our project in VM locally

You can also verify that these tests pass:

  - rust-hwlib: plucky/rust-hwlib/0.9.0~ppa8 [amd64]
    + ✅ rust-hwlib on plucky for amd64   @ 08.05.25 13:07:06
      • Log: https://autopkgtest.ubuntu.com/results/autopkgtest-plucky-nhutsko-testing/plucky/amd64/r/rust-hwlib/20250508_130706_1457d@/log.gz
    + ✅ rust-hwlib on plucky for amd64   @ 08.05.25 13:08:48
      • Log: https://autopkgtest.ubuntu.com/results/autopkgtest-plucky-nhutsko-testing/plucky/amd64/r/rust-hwlib/20250508_130848_46833@/log.gz


client/debian/control
```diff
@@ -41,8 +47,8 @@ Description: Library for querying Ubuntu hardware certification status - Rust so
 Package: hwctl
 Architecture: amd64
 Multi-Arch: foreign
-Depends: ${misc:Depends}, ${shlibs:Depends}
-Description: Command-line tool for checking Ubuntu hardware certification status - Rust source code
+Depends: kmod, lsb-release, ${misc:Depends}, ${shlibs:Depends}
+Description: CLI tool for checking Ubuntu hardware certification status - Rust source code
```
nadzyah, line 51: Has been shortened to fix "synopsis too long" lintian error

client/README.md
```diff
@@ -228,19 +228,38 @@ Not you can build the binary itself:
...
+for building the package.
 
 ```sh
-autopkgtest /path/to/<package>_<version>_source.changes -- lxd autopkgtest/ubuntu/plucky/amd64
+autopkgtest \
+  --apt-upgrade \
+  --shell-fail \
+  --output-dir dep8-rust-hwlib \
+  --setup-commands="mount -o remount,size=10G /tmp" \
+    /path/to/<package>_<version>_source.changes \
```
zyga, line 261: This can be . for the source tree. This would make the whole command copy paste friendly.

client/debian/control
```diff
@@ -12,7 +12,7 @@ Build-Depends: cargo:native (>= 1.75),
...
...macro2@1.0.95, pyo3-build-config@0.24.2, pyo3-ffi@0.24.2, pyo3-macros-backend@0.24.2, pyo3-macros@0.24.2, pyo3@0.24.2, quote@1.0.40, regex-automata@0.4.9, regex-syntax@0.8.5, regex@1.11.1, reqwest@0.12.15, ring@0.17.14, rustls-pemfile@2.2.0, rustls-pki-types@1.11.0, rustls-webpki@0.103.1, rustls@0.23.26, ryu@1.0.20, serde@1.0.219, serde_derive@1.0.219, serde_json@1.0.140, serde_urlencoded@0.7.1, shlex@1.3.0, simple_test_case@1.2.0, slab@0.4.9, smallvec@1.15.0, smbios-lib@0.9.2, socket2@0.5.9, stable_deref_trait@1.2.0, subtle@2.6.1, syn@2.0.100, sync_wrapper@1.0.2, synstructure@0.13.1, target-lexicon@0.13.2, time-core@0.1.4, time-macros@0.2.22, time@0.3.41, tinystr@0.7.6, tokio-macros@2.5.0, tokio-native-tls@0.3.1, tokio-rustls@0.26.2, tokio-util@0.7.15, tokio@1.45.0, tower-layer@0.3.3, tower-service@0.3.3, tower@0.5.2, tracing-core@0.1.33, tracing@0.1.41, try-lock@0.2.5, unicode-ident@1.0.18, unicode-width@0.1.14, unindent@0.2.4, untrusted@0.9.0, url@2.5.4, utf16_iter@1.0.5, utf8_iter@1.0.4, vcpkg@0.2.15, want@0.3.1, write16@1.0.0, writeable@0.5.5, yansi@1.0.1, yoke-derive@0.7.5, yoke@0.7.5, zerofrom-derive@0.1.6, zerofrom@0.1.6, zeroize@1.8.1, zerovec-derive@0.10.3, zerovec@0.10.4
```
zyga, line 15: Can you wrap-and-sort this please (run the tool)

client/debian/control
```diff
@@ -22,7 +22,13 @@ Rules-Requires-Root: no
 Package: librust-hwlib-dev
 Architecture: amd64
 Multi-Arch: foreign
-Depends: libssl-dev, pkgconf, python3, ${misc:Depends}, ${shlibs:Depends}
+Depends: kmod,
```
zyga, line 25: Does the devel package actually need those? Are they really invoked from tests (not mocked)?

client/debian/control
zyga: Perhaps drop the acronym. It reads well as-is

client/debian/tests/run-hwctl-tests.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+set -xeu
+
+python3 debian/tests/mock_server.py &
```
zyga, line 4: This should be systemd-run, you can then reliably stop it or wait for it.

client/debian/tests/run-hwctl-tests.sh
```diff
@@ -0,0 +1,33 @@
...
+SERVER_PID=$!
+sleep 1  # Give server time to start
```
zyga, line 6: The server should be ready when it signals readiness to systemd. This will race on slower hardware or virtual machines.

zyga (commented):
A few questions and a suggestion for making tests independent of the speed of the test machine.

client/README.md
nadzyah: Since this is a monorepo, just `.` may be confusing (also the changes file get generated in the parent dir). Or do you suggest mentioning that it should be done in the `client/` dir?

client/debian/control
nadzyah: I believe I already did that, but I’ve just checked that when I run the `wrap-and-sort` command, it doesn't change anything (but used to change the order of dependencies for example, so it works)

client/debian/control
nadzyah: The library uses these tools to collect OS information

client/README.md
zyga: Yeah, I meant the specific directory where it would work.

client/debian/control
zyga: Ah I expected it to make that long line different. Oh well.

client/debian/control
zyga: Yes but how does the developer package need them? Those should be on the non-devel package.

nadzyah: if you want to run the code that sends the certification status request, you need these packages

client/README.md
nadzyah: I actually get this error if I pass the dir name instead of source.chages file

```
The following packages have unmet dependencies:
 autopkgtest-satdep : Depends: librust-hwlib-dev but it is not installable
                      Depends: hwctl but it is not installable
E: Unable to correct problems, you have held broken packages.
E: The following information from --solver 3.0 may provide additional context:
   Unable to satisfy dependencies. Reached two conflicting decisions:
   1. autopkgtest-satdep:amd64=0 is selected for install
   2. autopkgtest-satdep:amd64 Depends librust-hwlib-dev
      but none of the choices are installable:
      [no choices]
```

client/debian/tests/run-hwctl-tests.sh
nadzyah: I've updated this with your suggestion. I've verified it locally in a VM that It works:

```
autopkgtest [22:58:54]: test rust-hwlib:hwctl: debian/tests/run-hwctl-tests.sh
autopkgtest [22:58:54]: test rust-hwlib:hwctl: [-----------------------
+ dirname debian/tests/run-hwctl-tests.sh
+ realpath debian/tests
+ systemd-run --unit=mock-hw-server --property=NotifyAccess=all --property=Type=notify --slice=background.slice --quiet python3 /tmp/autopkgtest.PCEAek/build.mGH/src/debian/tests/mock_server.py
+ systemctl is-active --quiet mock-hw-server.service
+ export HW_API_URL=http://localhost:8089
+ hwctl
+ OUTPUT={
  "status": "Not Seen"
}
+ EXPECTED_OUTPUT={ "status": "Not Seen" }
+ echo {
  "status": "Not Seen"
}
+ jq -S .
+ echo { "status": "Not Seen" }
+ jq -S .
+ diff -u /tmp/expected.json /tmp/out.json
+ systemctl stop mock-hw-server
Integration test passed successfully
+ echo Integration test passed successfully
autopkgtest [22:58:55]: test rust-hwlib:hwctl: -----------------------]
rust-hwlib:hwctl     PASS
autopkgtest [22:58:56]: test rust-hwlib:hwctl:  - - - - - - - - - - results - - - - - - - - - -
autopkgtest [22:58:56]: @@@@@@@@@@@@@@@@@@@@ summary
rust-hwlib:@         PASS
rust-hwlib:hwctl     PASS
```

client/debian/control
zyga: Right but who needs the -devel package? Someone bundling the library in their project (like hwctl). The devel package does not actually need them, right? You don't end up with a runtime dependency on -dev package.

client/debian/tests/run-hwctl-tests.sh
```diff
@@ -0,0 +1,42 @@
...
+set -xeu
+
+systemd-run --unit=mock-hw-server \
+            --property=NotifyAccess=all \
+            --property=Type=notify \
+            --slice=background.slice \
+            --quiet \
+            python3 "$(realpath "$(dirname "$0")")/mock_server.py"
+
+if ! systemctl is-active --quiet mock-hw-server.service; then
```
zyga, line 11: Doesn't systemd wait for the service to start before returning? What's the point of this part?

client/README.md
zyga: I don't know why but I don't think this is crucial.

zyga (approved):
LGTM, thank you for the improvements!

PR #312, opened 2025-05-14
[DRAFT] Replace dpkg with Rust ARCH const

pedro-avalos:
This removes the need for running the `dpkg` command in `hwctl`. Instead, we can use the `ARCH` constant that is provided by Rust.

I have, however, changed the `get_architecture` function to instead convert an arch string to a debian architecture string.

### DIFF
--- a/client/hwlib/src/collectors/os_info.rs
+++ b/client/hwlib/src/collectors/os_info.rs
@@ -22,7 +22,7 @@ use regex::Regex;
 use std::{fs::read_to_string, path::Path, process::Command};
 
 use crate::{
-    constants::{DPKG, LSB_RELEASE, LSMOD},
+    constants::{LSB_RELEASE, LSMOD},
     models::software::{KernelPackage, OS},
 };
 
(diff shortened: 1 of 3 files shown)


nadzyah:
We cannot do this because it doesn't match the way Checkbox gets this information
The ARCH variable is `x86_64` on amd64 devices, and `dpkg` returns `amd64` (what checkbox has as well)
This will break the package and the matching logic, so we should use the `dpkg` tool

client/hwlib/src/collectors/os_info.rs
```diff
@@ -84,9 +84,20 @@ impl KernelPackage {
...
+        "arm" => "armel",
+        "loongarch64" => "loong64",
+        "mips" => "mips",
+        "powerpc" => "powerpc",
+        "powerpc64" => "ppc64",
+        "riscv64" => "riscv64",
+        "x86" => "i386",
+        "x86_64" => "amd64",
+        _ => arch,
+    };
```
nadzyah, line 99: instead of supporting this mapping, let's just do the way checkbox does it and use `dpkg` command

PR #315, opened 2025-05-15
Replace lsb-release with os-release library

### LINKED ISSUE #314: Investigate `os-release` Rust library as an alternative to `lsb-release` executable
I think we should remove the dependency of the `lsb-release` executable, which requires us to spawn a process with a shell just to read some information that is in the filesystem. There is an `os-release` library already that is the backend to the `os-release` tool. This library would give us a simple Rust struct with the information we need, without the need for `lsb-release`.

TODO: Investigate if the `/etc/os-release` file that this depends on is available in all the OSes and versions we support

See: https://docs.rs/os-release/latest/os_release/index.html
See: https://www.linux.org/docs/man5/os-release.html
See: http://0pointer.de/blog/projects/os-release

pedro-avalos:
Resolves #314.
Resolves [CER-2974](https://warthogs.atlassian.net/browse/CER-2974).

This replaces the dependency of the executable `lsb-release` with the build-dependency of [`os-release`](https://docs.rs/os-release/latest/os_release/index.html) (a library under the Pop! OS org).

This would resolve some issues with packaging `hwctl` as a snap, since `lsb_release` is not usable by the snap (no interface provides access to it); there are other reasons, outlined below:

See: 

* [`checkbox-support` release info parsing](https://github.com/canonical/checkbox/blob/5c4aab054878fb67980799970907515ee7ef0b9a/checkbox-support/checkbox_support/helpers/release_info.py#L9)
* [Checkbox packaging metadata documentation](https://github.com/canonical/checkbox/blob/5c4aab054878fb67980799970907515ee7ef0b9a/docs/reference/units/packaging-meta-data.rst#L30)
* [Checkbox deprecation notice of using `lsb` --> `os` (read `/etc/os-release`)](https://github.com/canonical/checkbox/blob/5c4aab054878fb67980799970907515ee7ef0b9a/providers/resource/jobs/resource.pxu#L82)

Even the `lsb-release` package in Ubuntu actually reads the `os-release` file:

```
 ...
 This package contains a bare-bones implementation that uses the
 information in /etc/os-release instead of relying on LSB packages.
```

## TODO

- [x] Update and add tests.
- [x] Double-triple-quadruple check that `/etc/os-release` is everywhere we expect to package/distribute `hwctl` (it should be since systemd 245, which ships with focal)
- [x] Update AppArmor profile (updated in this branch, but not in AppArmor repo)

[CER-2974]: https://warthogs.atlassian.net/browse/CER-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nadzyah:
I'm unsure about this change because we need to collect the information the same way Checkbox does it, so it matches what we have in the reports. Checkbox doesn't parse `/etc/os-release` file (AFAIK) but uses the `lsb_release` command

pedro-avalos:
We do I think. I think I just need to change from using the `id` to the `name` field (lower case `ubuntu` vs `Ubuntu`)

client/debian/changelog
```diff
@@ -1,3 +1,9 @@
+rust-hwlib (0.9.2) questing; urgency=medium
+
+  * Replace lsb-release dependency with os-release Rust library
+
+ -- Pedro Avalos <pedro.avalosjimenez@canonical.com>  Sat, 17 May 2025 09:31:10 +0200
+
```
nadzyah, line 6: We shouldn't create a new release with this change, this should be handled later when we create a new release (moreover, 0.9.1 has not been published, we can do it with 0.9.1)

client/hwctl/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "hwctl"
 description = "hwlib backed CLI tool for collecting and sending hardware information to hwapi to find testing status of device models"
 license = "GPL-3.0-only"
-version = "0.9.1"
+version = "0.9.2"
```
nadzyah, line 5:
```suggestion
version = "0.9.1"
```

client/hwlib/Cargo.toml
```diff
@@ -2,7 +2,7 @@
 name = "hwlib"
 description = "Collects and sends hardware information to hwapi to find testing status of the device model"
 license = "GPL-3.0-only"
-version = "0.9.1"
+version = "0.9.2"
```
nadzyah, line 5:
```suggestion
version = "0.9.1"
```

client/hwlib/Cargo.toml
```diff
@@ -19,6 +19,7 @@ name = "hwlib"
 crate-type = [ "cdylib", "rlib",]
 
 [dependencies]
+os-release = { version = "0.1.0" }
```
nadzyah, line 22: since this is a new dependency, you need to update `debian/copyright` to mention its copyright and all its subdependencies' copyrights
you can check `Cargo.lock` changes to see which new dependencies (that are not filtered by cargo-vendor-filterer) have been added with this. Also, the copyright and license information can be fetched from a library  `Cargo.toml` file

client/hwlib/src/constants.rs
```diff
@@ -23,6 +23,8 @@ pub const PROC_CPUINFO_FILE_PATH: &str = "/proc/cpuinfo";
 pub const PROC_DEVICE_TREE_DIR_PATH: &str = "/proc/device-tree/";
 pub const PROC_VERSION_FILE_PATH: &str = "/proc/version";
 
+pub const OS_RELEASE_FILE_PATH: &str = "/etc/os-release";
+pub const SNAP_OS_RELEASE_FILE_PATH: &str = "/var/lib/snapd/hostfs/etc/os-release";
```
nadzyah, line 27: It feels kinda odd that the source code should be aware of the way it's packaged. are you sure this is the right way to handle it?

nadzyah (changes requested):

client/hwlib/src/models/request_validators.rs
```diff
@@ -130,7 +141,10 @@ impl CertificationStatusRequest {
         let vendor = system_info.manufacturer;
 
         let architecture = get_architecture(runner)?;
-        let os = OS::try_new(proc_version_filepath.as_path(), runner)?;
+
+        let os = OS::try_new(&os_release_filepath, &proc_version_filepath, runner)
+            .context("Failed to read OS release information")?;
```
nadzyah, line 146: we use this messages format across the project:

```suggestion
            .context("cannot read OS release information")?;
```

check also: https://github.com/canonical/hardware-api/pull/317#discussion_r2092933547

client/debian/changelog
nadzyah: Just add this point under the 0.9.1 release

client/hwlib/Cargo.toml
pedro-avalos: I've updated the copyright. There are no dependencies added apart from `os-release` (it has no subdependencies).

client/hwlib/src/models/request_validators.rs
pedro-avalos: Sounds good, I've updated it

client/hwlib/src/models/request_validators.rs
```diff
@@ -48,17 +48,27 @@ pub struct Paths {
...
+            PathBuf::from(constants::OS_RELEASE_FILE_PATH)
+        };
+
         Self {
             smbios_entry_filepath: PathBuf::from(smbioslib::SYS_ENTRY_FILE),
             smbios_table_filepath: PathBuf::from(smbioslib::SYS_TABLE_FILE),
             cpuinfo_filepath: PathBuf::from(constants::PROC_CPUINFO_FILE_PATH),
             max_cpu_frequency_filepath: PathBuf::from(constants::CPU_MAX_FREQ_FILE_PATH),
             device_tree_dirpath: PathBuf::from(constants::PROC_DEVICE_TREE_DIR_PATH),
+            os_release_filepath: os_release_filepath,
```
TheSignPainter98, line 71: In its current form, I think this line can be simplified. However, doing so will cause it to get visually lost among all the other lines which are responsible for both creating data and populating fields. If my above suggestion to use `Path` rather than `PathBuf` is accepted and if keeping `os_release_filepath` as a `PathBuf` is acceptable, then this issue is somewhat mitigated as this line could become
```rust
os_release_filepath: os_release_filepath.to_owned(),
```

This whole block is quite busy-looking now we're introducing a new line-style in, perhaps it'd be cleaner to refactor it? Please consider the [guidelines](https://canonical.github.io/rust-best-practices/code-discipline.html?highlight=struct#struct-population) around this :)

client/hwlib/src/collectors/os_info.rs
```diff
@@ -134,17 +107,10 @@ mod tests {
...
-        let mock_calls = vec![
-            ((LSB_RELEASE, vec!["-c"]), Ok("Codename: focal\n")),
-            ((LSB_RELEASE, vec!["-i"]), Ok("Distributor ID: Ubuntu\n")),
-            (
-                (LSB_RELEASE, vec!["-r"]),
-                Ok("No LSB modules are available.\nRelease: 20.04\n"),
-            ),
-            ((LSMOD, vec![]), Ok("Module Size Used\nsnd 61440 1\n")),
-        ];
+        let mock_calls = vec![((LSMOD, vec![]), Ok("Module Size Used\nsnd 61440 1\n"))];
```
TheSignPainter98, line 121: Feel free to ignore this, but for consistency with creating other empty containers, I think it would be nicer to use `Vec::new()`. This has two advantages:
1. When creating other empty containers, e.g. `HashMap`, `HashMap::new()` exists but `hashmap![]` does not
2. The docs provide us with a guarantee that `Vec::new()` does not allocate. The `vec![]` macro does not provide this guarantee

As these are tests, reducing allocations doesn't matter too much so going for the more concise option is also acceptable, but I see elsewhere in this codebase `Vec::new()` is already used

client/hwlib/src/collectors/os_info.rs
```diff
@@ -42,12 +41,15 @@ impl CommandRunner for SystemCommandRunner {
...
         runner: &impl CommandRunner,
     ) -> Result<Self> {
-        let codename = get_codename(runner)?;
-        let distributor = get_distributor(runner)?;
-        let version = get_version(runner)?;
+        let release = OsRelease::new_from(os_release_filepath)
+            .context("Failed to read OS release information")?;
+        let codename = release.version_codename;
+        let distributor = release.name;
+        let version = release.version_id;
```
TheSignPainter98, line 54: Would it be cleaner to pattern-match here?
```rust
let OsRelease {
    version_codename: codename,
    name: distributor,
    version_id: version,
    ..
} = release;
```

client/hwlib/src/models/request_validators.rs
```diff
@@ -48,17 +48,27 @@ pub struct Paths {
...
 
 impl Default for Paths {
     fn default() -> Self {
+        // Snap confinement exposes /etc/os-release in a different location
+        let snap_os_release_filepath = PathBuf::from(constants::SNAP_OS_RELEASE_FILE_PATH);
+        let os_release_filepath = if snap_os_release_filepath.exists() {
+            snap_os_release_filepath
+        } else {
+            PathBuf::from(constants::OS_RELEASE_FILE_PATH)
+        };
```
TheSignPainter98, line 63: This block looks surprising to me, do we need `hwlib` to be aware of two possible paths depending on how it's packaged? This feels very fragile to me, someone coming to this code in future may easily forget to check the values in this order, quietly leading to bugs. Instead, would be be cleaner to just have one `RELEASE_FILE_PATH` which is configured correctly at compile time?

client/hwlib/src/models/request_validators.rs
```diff
@@ -48,17 +48,27 @@ pub struct Paths {
...
     pub device_tree_dirpath: PathBuf,
+    pub os_release_filepath: PathBuf,
     pub proc_version_filepath: PathBuf,
 }
 
 impl Default for Paths {
     fn default() -> Self {
+        // Snap confinement exposes /etc/os-release in a different location
+        let snap_os_release_filepath = PathBuf::from(constants::SNAP_OS_RELEASE_FILE_PATH);
+        let os_release_filepath = if snap_os_release_filepath.exists() {
```
TheSignPainter98, line 59: There's something strange here, if `SNAP_OS_RELEASE_FILE_PATH` is inaccessible (e.g. due to incorrect permissions), `.exists()` will return `false` causing us to check for the info in the wrong file. When `hwlib` is in a snap, this will trigger errors referring to the wrong file!

TheSignPainter98 (changes requested):
Replacing lsb-release with os-release seems like an excellent idea. There's some fragile logic around path handling which I think we should be careful about, other than that please consider the following tweaks

client/hwlib/src/collectors/os_info.rs
```diff
@@ -42,12 +41,15 @@ impl CommandRunner for SystemCommandRunner {
...
     pub(crate) fn try_new(
+        os_release_filepath: &Path,
         proc_version_filepath: &Path,
         runner: &impl CommandRunner,
     ) -> Result<Self> {
-        let codename = get_codename(runner)?;
-        let distributor = get_distributor(runner)?;
-        let version = get_version(runner)?;
+        let release = OsRelease::new_from(os_release_filepath)
+            .context("Failed to read OS release information")?;
```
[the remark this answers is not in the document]
pedro-avalos: Ah yes Nadzeya brought that up, thanks for catching that

client/hwlib/src/collectors/os_info.rs
pedro-avalos: Still newer to and learning Rust, didn't think of that, that's pretty cool!

client/hwlib/src/constants.rs
pedro-avalos: I've moved it to a build environment variable; I didn't know about these. Let me know if my `build.rs` is missing something, though

client/hwlib/src/models/request_validators.rs
pedro-avalos: Since the path is now a constant like the others, I've moved it with the rest of the `PathBuf::from`, or should I move all of these outside the block?

client/hwlib/src/models/request_validators.rs
pedro-avalos: This is now a build env var

client/hwlib/build.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    let os_release_file_path =
+        std::env::var("OS_RELEASE_FILE_PATH").unwrap_or("/etc/os-release".to_string());
+    println!(
+        "cargo:rustc-env=OS_RELEASE_FILE_PATH={}",
+        os_release_file_path
+    );
+}
```
nadzyah, line 8: are you sure we need this logic since this file should always be `/etc/os_release`?
for tests we can use the same logic as for other paths

client/hwlib/src/constants.rs
```diff
@@ -21,8 +21,8 @@ pub const CERT_STATUS_ENDPOINT: &str = "/v1/certification/status";
 pub const CPU_MAX_FREQ_FILE_PATH: &str = "/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq";
 pub const PROC_CPUINFO_FILE_PATH: &str = "/proc/cpuinfo";
 pub const PROC_DEVICE_TREE_DIR_PATH: &str = "/proc/device-tree/";
+pub const OS_RELEASE_FILE_PATH: &str = env!("OS_RELEASE_FILE_PATH", "/etc/os-release");
```
nadzyah, line 24: I suggest being consistent with other paths and specify `/ets/os-release`
Everything that can be customised by an end-user needs an additional security review and can be a potential attack surface. it's better to avoid this if we can :)

client/snapcraft.yaml
```diff
@@ -39,3 +39,5 @@ parts:
       metadata="$(cargo metadata --no-deps --format-version 1)"
       version="$(echo "$metadata" | jq -r .packages[0].version)"
       craftctl set version="$version"
+    build-environment:
+      - OS_RELEASE_FILE_PATH: /var/lib/snapd/hostfs/etc/os-release
```
nadzyah, line 43: aaaaaaah, that's why you're using an env variable. Eh, are you sure there is no other way? :(

pedro-avalos: Yeah, it's a build/compile-time variable now if I understood @TheSignPainter98's suggestion correctly.

client/hwlib/src/constants.rs
pedro-avalos: It's customized at build-time, not by the user

nadzyah: Please add a comment explaining this for the Security team

pedro-avalos: Done

client/hwlib/src/constants.rs
```diff
@@ -23,6 +23,9 @@ pub const PROC_CPUINFO_FILE_PATH: &str = "/proc/cpuinfo";
 pub const PROC_DEVICE_TREE_DIR_PATH: &str = "/proc/device-tree/";
 pub const PROC_VERSION_FILE_PATH: &str = "/proc/version";
 
+// The os-release file path is set by a build-time environment variable
```
nadzyah, line 26:
```suggestion
// The os-release file path is set during the build time
```

I think it’s important to indicate that it cannot be modified by a user (cannot be changed during the runtime), but feel free to ignore id you think that’s fine

nadzyah (approved):
Thank you SO MUCH for taking care of this! Your work shines! There is only one comment; feel free to ignore it. Also, please make sure Ed approves it before you merge it

client/hwlib/src/constants.rs
pedro-avalos: Sure, that's a reasonable change, and makes the comment more definitive

client/hwlib/build.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    let os_release_file_path =
+        std::env::var("OS_RELEASE_FILE_PATH").unwrap_or("/etc/os-release".to_string());
+    println!(
+        "cargo:rustc-env=OS_RELEASE_FILE_PATH={}",
+        os_release_file_path
+    );
```
TheSignPainter98, line 7: Just for clarity, I think `os_release_file_path` can be put in the curly braces
```rust
"cargo:rustc-...={os_release_file_path}"
```
Also, does `rustfmt` break the line here? If possible, I think it could be nice to keep this `println!` call to just one line

client/hwlib/src/collectors/os_info.rs
```diff
@@ -42,12 +41,19 @@ impl CommandRunner for SystemCommandRunner {
...
-        let version = get_version(runner)?;
+        let release = OsRelease::new_from(os_release_filepath).with_context(|| {
+            format!("cannot read OS release information from: {os_release_filepath:?}",)
+        })?;
+        let OsRelease {
+            version_codename: codename,
+            name: distributor,
+            version_id: version,
+            ..
+        } = release;
```
TheSignPainter98, line 56: I think the two new `let` statements form a block of strongly-associated code, hence a newline just here could be beneficial

client/hwlib/src/constants.rs
```diff
@@ -23,6 +23,9 @@ pub const PROC_CPUINFO_FILE_PATH: &str = "/proc/cpuinfo";
 pub const PROC_DEVICE_TREE_DIR_PATH: &str = "/proc/device-tree/";
 pub const PROC_VERSION_FILE_PATH: &str = "/proc/version";
 
+// The os-release file path is set during the build time
+// to allow for snap packaging, which exposes the file in a different location.
+pub const OS_RELEASE_FILE_PATH: &str = env!("OS_RELEASE_FILE_PATH", "/etc/os-release");
```
TheSignPainter98, line 28: Is there a little code-duplication here? I think the fallback to `/etc/os-release` is already handled by `build.rs`---if something changes and `build.rs` no longer defines the env var we expect, I think it could be nice to trigger a compilation error here (i.e could the fallback on this line be removed?)

client/hwlib/src/constants.rs
```diff
@@ -23,6 +23,9 @@ pub const PROC_CPUINFO_FILE_PATH: &str = "/proc/cpuinfo";
 pub const PROC_DEVICE_TREE_DIR_PATH: &str = "/proc/device-tree/";
 pub const PROC_VERSION_FILE_PATH: &str = "/proc/version";
 
+// The os-release file path is set during the build time
+// to allow for snap packaging, which exposes the file in a different location.
```
TheSignPainter98, line 27: I think the second line here is slightly leaky---there may be other reasons in future unrelated to snaps which require us to use `env!` which could cause this comment to quietly go out of date. I think it may be cleaner to just remove the second line here.

The first line is also slightly redundant, but given that this is a non-standard environment variable, I see no harm in keeping it. Small nitpick, I think it would be more natural to say that the '`os-release` file path is _defined_ in `build.rs`' (rather than 'set')

client/hwlib/build.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    let os_release_file_path =
+        std::env::var("OS_RELEASE_FILE_PATH").unwrap_or("/etc/os-release".to_string());
```
TheSignPainter98, line 3: Small nitpick here, the `.to_string()` method comes from the `Display` trait which allows for arbitrary, pretty re-formatting. I think the intention here isn't pretty-printing, but rather to change the ownership model from borrowed (`&str`) to owned (`String`), hence using the `.to_owned()` method would be slightly better to signal the intention

TheSignPainter98 (changes requested):
Thanks for the changes, this is looking really nice and clean now. Just a few small notes then I think this is ready

client/hwlib/build.rs
pedro-avalos: I now put it in the curly braces now and `rustfmt` didn't break the line

client/hwlib/src/constants.rs
```diff
@@ -23,6 +23,8 @@ pub const PROC_CPUINFO_FILE_PATH: &str = "/proc/cpuinfo";
 pub const PROC_DEVICE_TREE_DIR_PATH: &str = "/proc/device-tree/";
 pub const PROC_VERSION_FILE_PATH: &str = "/proc/version";
 
+// The os-release file path is defined in build.rs at build time
```
TheSignPainter98, line 26: Full stop, please ;)

TheSignPainter98 (approved):
This is excellent, thanks for the tweaks!

PR #316, opened 2025-05-16
Add AppArmor profile and autopkgtest for it

nadzyah:
This PR adds the AppArmor profile, merged [in this PR](https://gitlab.com/apparmor/apparmor/-/merge_requests/1658) and makes sure it gets installed. I've tested it locally using the instructions from the client README.
It also updates the changelog to match what's in the LP package changelog.

Since version 0.9.0 is in the universe queue, we're releasing 0.9.1 version here


renanrodrigo (approved):
Tested, built, autopkgtests are good, LGTM

PR #318, opened 2025-05-16
Remove librust-hwlib-dev from debian source package

nadzyah:
It was discussed with the archive-admins, the pro-client team, and the MIR reviewer that having the `librust-hwlib-dev` file is not useful and even harmful to have in the Ubuntu main archive.
Since the Ubuntu Rust ecosystem doesn't follow traditional dependency management for crates, so even another Rust library needs to use it, they'll have to vendor it and won't be able to use the Debian package.
`pro-client` won't be able to use it as well since it's not a Python package. As it was decided, we should not install this library and keep only the binary package.


renanrodrigo (approved):
I have built and tested this, - hwctl installs fine.
Found #319 and #320 while testing, but could confirm this changeset is not responsible for the errors found.

PR #322, opened 2025-05-19
Update client snapcraft.yaml

pedro-avalos:
The client snap was missing a homepage, license, specification, and stage packages.

Since we are dropping `dpkg` (#323) and `lsb-release` (#315), I only added `kmod` as a stage dependency.

This PR also updates the packaging workflow to:

1. Upload snapcraft logs on failure
2. Use [canonical/snapcraft-multiarch-action](https://github.com/canonical/snapcraft-multiarch-action/) to build the snap; this will allow us to easily add new architectures in the future.


client/snapcraft.yaml
```diff
@@ -1,11 +1,22 @@
 name: hwctl
 adopt-info: hwctl
+license: GPL-3.0-only
 summary: Canonical Certified Hardware API Client
 description: |
-  hwctl is a hwlib backed CLI tool for collecting and sending hardware
-  information to hwapi to find testing status of device models
+  `hwctl` is a `hwlib` backed CLI tool for collecting and sending hardware
+  information to a Hardware API server to find testing status of device models.
```
nadzyah, line 7:
```suggestion
  information to a Hardware API server to find certification status of device models.
```

At least this is what we’re saying in a Debian package description

nadzyah (approved):

client/snapcraft.yaml
pedro-avalos: Let's be consistent then, I've changed it

PR #323, opened 2025-05-19
Replace dpkg with Rust ARCH

### LINKED ISSUE #321: `dpkg` is not available within snap confinement
None of the snapd interfaces allow access to the `dpkg` executable. We need to either:

- get `dpkg-architecture` / `dpkg` into a snapd interface (a custom one for `hwctl` or a builtin one)
- replace the use of `dpkg` with some other method to get the architecture (and convert it to the [Checkbox-supported list of architectures](https://github.com/canonical/checkbox/blob/5c4aab054878fb67980799970907515ee7ef0b9a/submission-schema/schema.json#L920))

EDIT: After discussing with George, this seems to be a more important change than initially thought. If we want to run `hwctl` in Core systems, we can't rely on `dpkg`. I think the Rust constant [`ARCH`](https://doc.rust-lang.org/std/env/consts/constant.ARCH.html) should be as good as using `uname -m` or `/proc/sys/kernel/arch`. We would then need to convert that to the accepted Checkbox choices.

pedro-avalos:
Resolves #321.
Resolves [CER-2973](https://warthogs.atlassian.net/browse/CER-2973).

As stated in the linked issue, we can't really rely on `dpkg`, as that does not work in all environments, especially within Snap confinement or in a Core system.

`ARCH` is a Rust constant that provides the architecture string. We can parse that and convert it to [a debian architecture supported by Checkbox](https://github.com/canonical/checkbox/blob/5c4aab054878fb67980799970907515ee7ef0b9a/submission-schema/schema.json#L920).

## TODO

- [X] Update unit tests
- [x] Update AppArmor profile (needs verifying that it doesn't need anything more)

[CER-2973]: https://warthogs.atlassian.net/browse/CER-2973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


client/debian/changelog
```diff
@@ -1,3 +1,9 @@
+rust-hwlib (0.9.2) questing; urgency=medium
+
+  * Replace dpkg command with arch (coreutils)
+
+ -- Pedro Avalos <pedro.avalosjimenez@canonical.com>  Mon, 19 May 2025 13:26:35 -0400
+
```
nadzyah, line 6: let's fix it with v0.9.1

client/debian/usr.bin.hwctl
```diff
@@ -32,14 +32,13 @@ profile hwctl /usr/bin/hwctl {
   @{PROC}/@{pid}/cgroup r,
 
   # for collecting OS information
-  /usr/bin/{dpkg,kmod} cx,
+  /usr/bin/{arch,kmod} cx,
```
nadzyah, line 35: please check if we should remove the `/sys/fs/cgroup/**/cpu.max r,` line above

client/hwlib/src/models/request_validators.rs
```diff
@@ -255,7 +255,7 @@ mod tests {
             .collect::<String>();
 
         let mock_calls = vec![
-            ((constants::DPKG, vec!["--print-architecture"]), Ok("amd64")),
+            ((constants::ARCH, vec![]), Ok("x86_64")),
```
nadzyah, line 258: the version has to be `amd64` as all our checkbox submissions use it

nadzyah (changes requested):

client/hwlib/src/models/request_validators.rs
nadzyah: In this case the test should call the conversion function because the final result has to be “amd64”

client/debian/usr.bin.hwctl
pedro-avalos: Hm, it seems to still get that AppArmor denial, but the program does not crash/error at the architecture step. I'd say we leave the `cpu.max` files in AppArmor because we are accessing them, though perhaps not using the information?

```console
$ sudo hwctl
ERROR: cannot collect system data

Caused by:
    0: cannot get release info using lsb_release {flag}
    1: No such file or directory (os error 2)
```

client/hwlib/src/collectors/os_info.rs
```diff
@@ -91,11 +91,17 @@ impl KernelPackage {
     }
 }
 
-pub(crate) fn get_architecture(runner: &impl CommandRunner) -> Result<String> {
-    let arch = runner
-        .run_command(DPKG, &["--print-architecture"])
-        .with_context(|| "cannot determine system architecture")?;
-    Ok(arch.trim().to_owned())
+pub(crate) fn parse_debian_architecture(arch: &str) -> Result<String> {
```
pedro-avalos, line 94: Let me know what you think, but this function now doesn't raise any exceptions. Perhaps either:

1. I should remove the `Result<...>` portion of the function return type
2. Specify the full conversion mapping to the [Checkbox supported architectures](https://github.com/canonical/checkbox/blob/5c4aab054878fb67980799970907515ee7ef0b9a/submission-schema/schema.json#L920) and raise errors on any other architecture

nadzyah: I think the function name can remain `get_architecture` because it's confusing why it's a debian architecture (it's actually a machine arch :D)

pedro-avalos: Hm, but it is not "getting" the architecture; it is parsing out an architecture identifier from a given architecture string.

pedro-avalos: Well, that's the debian architecture identifier that we use in Checkbox and elsewhere.

pedro-avalos: Or ["debian architecture name"](https://wiki.debian.org/Multiarch/Tuples#Architectures_in_Debian)

client/hwlib/src/collectors/os_info.rs
```diff
@@ -91,11 +91,17 @@ impl KernelPackage {
...
+    let deb_arch = match arch.trim() {
+        "aarch64" => "arm64",
+        "arm" => "armhf",
+        "loongarch64" => "loong64",
+        "powerpc64" => "ppc64el",
+        "x86" => "i386",
+        "x86_64" => "amd64",
+        _ => arch.trim(),
+    };
```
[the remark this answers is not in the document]
pedro-avalos: the `riscv` architecture name is the same as the debian architecture name, so it falls under the `_` case, which just returns the trimmed architecture name :wink:

client/hwlib/src/collectors/os_info.rs
```diff
@@ -137,10 +143,9 @@ mod tests {
     use crate::helpers::test_utils::{get_test_filepath, MockCommandRunner};
 
     #[test]
-    fn test_get_architecture() {
-        let mock_calls = vec![((DPKG, vec!["--print-architecture"]), Ok("amd64\n"))];
-        let mock_runner = MockCommandRunner::new(mock_calls);
-        let result = get_architecture(&mock_runner);
+    fn test_parse_debian_architecture() {
+        let arch = "x86_64";
+        let result = parse_debian_architecture(&arch);
```
TheSignPainter98, line 148: Is this `&` here necessary? I think clippy may have a thing or two to say :P Also, if `arch` is only used once, I think it's okay to drop the `let arch =` line

TheSignPainter98 (changes requested):
Looks great! Just some trivials

client/hwlib/src/collectors/os_info.rs
pedro-avalos: I've dropped the `let arch =` line and put `"x86_64"` inside the function call, you're right

client/hwlib/src/collectors/os_info.rs
```diff
@@ -91,11 +91,17 @@ impl KernelPackage {
...
+    let deb_arch = match arch.trim() {
+        "aarch64" => "arm64",
+        "arm" => "armhf",
+        "loongarch64" => "loong64",
+        "powerpc64" => "ppc64el",
+        "x86" => "i386",
+        "x86_64" => "amd64",
+        _ => arch.trim(),
+    };
+    Ok(deb_arch.to_owned())
```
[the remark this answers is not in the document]
pedro-avalos: Changed the return type to `Result<&str>`

client/hwlib/src/collectors/os_info.rs
```diff
@@ -91,11 +91,26 @@ impl KernelPackage {
...
 }
 
-pub(crate) fn get_architecture(runner: &impl CommandRunner) -> Result<String> {
-    let arch = runner
-        .run_command(DPKG, &["--print-architecture"])
-        .with_context(|| "cannot determine system architecture")?;
-    Ok(arch.trim().to_owned())
+pub(crate) fn to_debian_architecture(arch: &str) -> Result<&str> {
+    match arch.trim() {
+        "aarch64" => Ok("arm64"),
```
TheSignPainter98, line 96: The many `Ok` introduce a lot of noise here, I think it would be cleaner to re-introduce the `let` binding and explicitly return in the error case, i.e.
```rust
let ret = match arch.trim() {
    "arm" => "armhf",
    ...
    _ => return Err(...), 
};
Ok(ret)
```

client/hwlib/src/collectors/os_info.rs
```diff
@@ -91,11 +91,26 @@ impl KernelPackage {
...
+        "mips64r6" => Ok("mips64r6"),
+        "powerpc" => Ok("powerpc"),
+        "powerpc64" => Ok("ppc64el"),
+        "riscv64" => Ok("riscv64"),
+        "s390x" => Ok("s390x"),
+        "sparc" => Ok("sparc"),
+        "sparc64" => Ok("sparc64"),
+        "x86" => Ok("i386"),
+        "x86_64" => Ok("amd64"),
+        _ => Err(anyhow!("cannot convert to debian architecture: {arch:?}")),
```
TheSignPainter98, line 112: The exhaustiveness is looking nice, I'm not sure about this error form though, perhaps this error message would read slightly more naturally if the `{arch:?}` were in the middle? As in
```rust
"cannot convert {arch:?} to debian architecture"
```

client/hwlib/src/models/request_validators.rs
```diff
@@ -129,7 +130,9 @@ impl CertificationStatusRequest {
         let model = system_info.product_name;
         let vendor = system_info.manufacturer;
 
-        let architecture = get_architecture(runner)?;
+        let architecture = to_debian_architecture(ARCH)
+            .with_context(|| format!("cannot parse architecture: {ARCH:?}"))?
```
TheSignPainter98, line 134: I think we should be careful about the use of colons in error messages, here. I think after a colon, the  reader can expect to see some inner reason for the error, but here, it's just the data which caused the error. Perhaps it would  be cleaner to drop the colon after architecture? If this is an existing convention across the codebase, feel free to address this in another PR

TheSignPainter98 (approved):
This is coming along nicely, thanks for the tweaks! Here are a few trivial points, please consider these before merging. I'll be OoO for the next few days, so I'm pre-approving to unblock you

PR #324, opened 2025-05-20
Fix BIOS date parsing in hwlib

### LINKED ISSUE #320: Error running `hwctl` on a questing lxd VM
While trying to run the hwctl command in a questing VM, I get the following error:

```
root@questing-test~# hwctl
ERROR: cannot collect system data

Caused by:
    0: the 'month' component could not be parsed
    1: the 'month' component could not be parsed
```

- Running on Questing Quokka (25.10)
- hwctl version v0.9.1

nadzyah:
Closes #320 

The problem was that usually BIOS release date is formatted using double digits for a month and a day (like `01/02/2023`). However, in the case of questing VM image, it was formatted using single digits like `1/2/2023`.
This PR fixes the date parsing for the BIOS release. I've also tested it locally and in questing VM to verify that it works.

Here are the steps describing how a reviewer can verify the changes:

1. Build the cargo binary
```bash
cd client/
cargo build
```
2. Launch the questing VM image 
```bash
lxc launch ubuntu-daily:25.10 questing-vm -p default --vm
lxc file push ./target/debug/hwctl questing-vm/home/ubuntu/
```
3. Run the command:
```bash
lxc exec questing-vm -- su - ubuntu
sudo ./hwctl
```


renanrodrigo:
@nadzyah thanks for taking a look!
I downloaded the artifacts from this PR to a Questing VM, and I'm getting a different error now:

```
root@questing-test:~# hwctl 
ERROR: cannot collect system data

Caused by:
    failed to capture information using regex
```

renanrodrigo:
Alright, makes sense! Can be fixed in the later changeset then

renanrodrigo (approved):

client/hwlib/src/helpers.rs
```diff
@@ -16,14 +16,30 @@
...
 pub(crate) fn append_to_pathbuf(p: PathBuf, s: &str) -> PathBuf {
     let mut p = p.into_os_string();
     p.push(s);
     p.into()
 }
 
+pub(crate) fn normalize_date(date_str: &str) -> Result<Date> {
+    let parts: Vec<_> = date_str.trim().split('/').collect();
+    if parts.len() != 3 {
+        return Err(anyhow!("invalid date format: {date_str}"));
```
TheSignPainter98, line 32: This is very pedantic, but I think this error occurs if `date_str` is invalidly formatted, rather than `date_str` itself being an invalid format (`date_str != "mm-dd-yyyy"`)

TheSignPainter98 (changes requested):
The logic looks good, just some minor notes on wording and cleanliness

client/hwlib/src/helpers.rs
```diff
@@ -16,14 +16,30 @@
...
+    }
```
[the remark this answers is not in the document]
nadzyah: your suggestion fails with this error:

```
error[E0277]: the trait bound `Vec<&str>: StdError` is not satisfied
   --> hwlib/src/helpers.rs:34:10
    |
34  |         .with_context(|| format!("cannot parse date {date_str:?}"))?;
    |          ^^^^^^^^^^^^ the trait `StdError` is not implemented for `Vec<&str>`
    |
    = help: the trait `anyhow::context::private::Sealed` is implemented for `Result<T, E>`
    = note: required for `Vec<&str>` to implement `anyhow::context::ext::StdError`
    = note: required for `Result<[&str; 3], Vec<&str>>` to implement `anyhow::context::private::Sealed`
note: required by a bound in `with_context`
   --> /home/nadzeya/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/anyhow-1.0.98/src/lib.rs:619:26
    |
619 | pub trait Context<T, E>: context::private::Sealed {
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Context::with_context`
...
627 |     fn with_context<C, F>(self, f: F) -> Result<T, Error>
    |        ------------ required by a bound in this associated function
```


So I have to change it this way

```rust
let [raw_month, raw_day, raw_year]: [&str; 3] = date_str
        .trim()
        .split('/')
        .collect::<Vec<_>>()
        .try_into()
        .map_err(|_| anyhow!("expected 3 components in date"))
        .with_context(|| format!("cannot parse date {:?}", date_str))?;
```

client/hwlib/src/helpers.rs
nadzyah: yes, I've just checked and it fails with  `cannot parse date "2-2-2022"`

client/hwlib/src/helpers.rs
```diff
@@ -16,14 +16,48 @@
...
+            let yy = raw_year
+                .parse::<i32>()
+                .context(format!("cannot parse year: {raw_year}"))?;
+            1900 + yy
+        }
+        4 => raw_year
+            .parse::<i32>()
+            .context(format!("cannot parse year: {raw_year}"))?,
+        _ => return Err(anyhow!(format!("invalid year format: {raw_year}"))),
+    };
```
nadzyah, line 55: Added the case for handling different year lengths according to the SMBIOS standard

client/hwlib/src/helpers.rs
```diff
@@ -16,14 +16,48 @@
...
+use time::Date;
 
 pub(crate) fn append_to_pathbuf(p: PathBuf, s: &str) -> PathBuf {
     let mut p = p.into_os_string();
     p.push(s);
     p.into()
 }
 
+pub(crate) fn parse_date(raw_date: &str) -> Result<Date> {
+    let [raw_month, raw_day, raw_year] = raw_date
```
TheSignPainter98, line 30: I wondered whether it would be possible to simplify this parsing exercise by using `time::format_description!` to allow the `time` crate to do the heavy lifting for us, but it looks like it doesn't support the single-digit day/month format  we need. This PR's approach looks good

client/hwlib/src/helpers.rs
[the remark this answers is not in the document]
TheSignPainter98: Ah! I didn't realise the `try_into` implementation here returned the original `Vec`, I thought it would have been some error, nice. The above looks good to me

client/hwlib/src/helpers.rs
```diff
@@ -16,14 +16,48 @@
...
 use std::path::PathBuf;
+use time::Date;
 
 pub(crate) fn append_to_pathbuf(p: PathBuf, s: &str) -> PathBuf {
     let mut p = p.into_os_string();
     p.push(s);
     p.into()
 }
 
+pub(crate) fn parse_date(raw_date: &str) -> Result<Date> {
```
TheSignPainter98, line 29: I see that the date in `bios.release_date` is formatted as `yyyy-mm-dd`, but this function only handles dates of a slightly different form. Perhaps to make it clearer which places this function should and shouldn't be used, it could be called something like  `parse_smbios_date`?

TheSignPainter98 (approved):
This is looking nice and clean now, thanks for the changes! I'm pre-approving pending a couple of trivials, feel free to merge once considered :)

client/hwlib/src/helpers.rs
TheSignPainter98: Ah, thanks for the context !

client/hwlib/src/helpers.rs
pedro-avalos: Hm, just looking at the `time` crate. Am I misunderstanding this, or would it actually be able to parse the single-digit day/month format?

> The `padding` modifier, such as used on the `hour` and `month` components can be padded with `space`, `zero` or `none`. In the case of `none` leading zeros are still accepted during parsing.

https://time-rs.github.io/book/api/format-description.html#modifiers

nadzyah:
I've decided that this logic should be part of the server implementation, and the client can send it just as a raw string. Currently, the server doesn't even use this information, so the only change was to accept it as a string.

PR #325, opened 2025-05-21
Extract shared worspace data

pedro-avalos:
This PR extracts shared information between the library and application. Namely, this extracts:

- package metadata (version, authors, links, etc.)
- shared dependencies (allows us to define shared dependencies, versions, and default features)

This allows us to reduce duplication while also reducing the number of places to edit when bumping the version or updating/changing dependencies.

See: https://doc.rust-lang.org/cargo/reference/workspaces.html


client/Cargo.toml
```diff
@@ -1,10 +1,25 @@
...
-    "hwctl",
-    "hwlib"
-]
+members = ["hwctl", "hwlib"]
 resolver = "2"
-exclude = ["rust-vendor",]
+exclude = ["rust-vendor"]
+
+[workspace.package]
+license = "GPL-3.0-only"
```
nadzyah, line 7: We're using LGPLv3 for hwlib and GPLv3 for hwctl

client/Cargo.toml
```diff
@@ -1,10 +1,25 @@
 [workspace]
-members = [
-    "hwctl",
-    "hwlib"
-]
+members = ["hwctl", "hwlib"]
 resolver = "2"
-exclude = ["rust-vendor",]
+exclude = ["rust-vendor"]
```
nadzyah, line 4: We should also exclude the debian folder (I forgot to move it from hwlib/Cargo.toml

client/Cargo.toml
pedro-avalos: Ah, didn't notice, that one needs to be independent for each, then

nadzyah:
Please resolve the conflicts, and I'll approve it

PR #326, opened 2025-05-22
Remove the usage of docker mirror in CI

nadzyah:
Since the runners are getting moved to PS7, we need to change the setup and use the `DOCKERHUB_MIRROR` environment variable provided by a runner. The default value is specified in case a job gets picked up by a PS6 runner, which may not provide such variable

UPD: The usage of the mirror has been dropped.


omar-selo (approved):
Looks good!

PR #334, opened 2025-05-27
Match system category when looking for motherboard

nadzyah:
Turned out that sometimes motherboard is classified as `SYSTEM` device (for example here: https://ubuntu.com/certified/202006-27980)
The server matching logic is being updated here to handle this case.


omar-selo (changes requested):
Looks good, but it would be great if we add test cases for this

omar-selo (approved):

PR #384, opened 2025-07-07
fix: remove git tag dependency in hwlib version validation

nadzyah:
The `.github/workflows/test_hwlib_debian_build.yaml` requires a git tag to exist before the PR can be merged. However, the git tag needs to be created after the PR was merged (with a new release). This creates a situation where neither can happen first.

The Solution:
- Remove the git tag dependency entirely
- Only validate that the versions in `Cargo.toml` and `debian/changelog` are consistent with each other
- This ensures version consistency without requiring external git tags


pedro-avalos (approved):
LGTM

PR #390, opened 2025-07-14
fix: update apparmor rm mask

nadzyah:
While the behaviour and the meaning of the masks remain the same, it just becomes less confusing and doesn't look like `rm` (removal)
The same syntax is used for the profile bin: `/usr/bin/hwctl mr,`


Hook25 (approved):
+1

PR #464, opened 2025-12-05
ci: Improve server testability

pedro-avalos:
This PR is an overhaul of the way we test the server. The following changes are noteworthy:

- Add Ruff rules (we weren't really enforcing any rules apart from the very bare minimum that Ruff includes by default)
  - Added isort, pylint, flake8 (builtins, bugbear, and bandit); these are what are used in TICS (which will be added in a follow-up PR) and we should be checking against them
  - Resolved any issues that Ruff brought up as a result of these changes
- Add `tox.ini` configuration to format, lint, and run unit/coverage tests
- Replace the Docker setup for testing with the tox setup

### DIFF
--- a/.github/workflows/test_server.yaml
+++ b/.github/workflows/test_server.yaml
@@ -25,25 +25,20 @@ jobs:
       run:
         working-directory: server
     steps:
-      - name: Checkout repo & submodules
+      - name: Checkout
         uses: actions/checkout@v6
         with:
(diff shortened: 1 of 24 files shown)


nadzyah:
With this quality of commit structure and messages, we can even configure to keep the linear history in main instead of squashes

nadzyah (approved):

PR #562, opened 2026-06-11
hwclt: Add cache for requests

sergio-costas:
This PR adds a cache to preserve the latest state, and defines the cases where the cache has to be invalidated and the Certification Status must be requested again.

By default, the cache is stored at $SNAP_DATA, since the idea is to go full-snap.

- Unit tests
- Manual testing steps:

Built it and tested both "as-is" and with "sudo", both without parameters and with a wrong "--server" parameter. The first one returned, as expected, an error because it can't collect the data. The second succeeds. The third fails, as expected, with an error because the server doesn't return the expected data format.

Now it always returns a JSON through STDOUT, but if there is an error, it will also return it through STDERR as before.

## Checklist

Still require more unitary tests.

- [X] I have added necessary tests.
- [ ] I have added or updated any relevant documentation (if needed).

## Additional Notes

[contribution guidelines]: https://github.com/canonical/hardware-api/blob/main/CONTRIBUTING.md
[cla]: https://ubuntu.com/legal/contributors

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -24,6 +24,7 @@ required-features = ["cli"]
 
 [dependencies]
 anyhow = "~1.0.0"
+chrono = "0.4.45"
 clap = { version = "4.5.41", features = ["derive", "env"], optional = true }
 minreq = { version = "3.0.0", features = [
     "native-tls",
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -35,11 +35,11 @@ pyo3 = { version = "0.29.0", features = ["extension-module"], optional = true }
 serde = { version = "~1.0.0", features = ["derive"] }
 serde_json = "~1.0.0"
 smbios-lib = "0.9.0"
-test-temp-dir = "0.7.0"
 
 [dev-dependencies]
 pretty_assertions = "1.4.1"
 simple_test_case = "1.2.0"
+test-temp-dir = "0.7.0"
 
 [features]
 cli = ["clap"]
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -35,6 +35,7 @@ pyo3 = { version = "0.29.0", features = ["extension-module"], optional = true }
 serde = { version = "~1.0.0", features = ["derive"] }
 serde_json = "~1.0.0"
 smbios-lib = "0.9.0"
+test-temp-dir = "0.7.0"
 
 [dev-dependencies]
 pretty_assertions = "1.4.1"
--- a/client/bin/hwctl.rs
+++ b/client/bin/hwctl.rs
@@ -63,7 +63,7 @@ fn run(server_url: String) -> Result<()> {
 
     let response = check_certification_status(
         server_url,
-        hwlib::CheckCertificationMode::Normal,
+        hwlib::CheckCertificationMode::Forced,
         &current_hardware,
         None,
     )
@@ -73,5 +73,13 @@ fn run(server_url: String) -> Result<()> {
         serde_json::to_string_pretty(&response).context("cannot serialize response as JSON")?;
     println!("{response_json}");
 
-    Ok(())
+    let (staled, stale_reason) = response.is_staled();
+
+    if staled {
+        return Err(anyhow::anyhow!(
+            stale_reason.unwrap_or("unknown reason".to_string())
+        ));
+    } else {
+        return Ok(());
+    }
 }
--- a/client/bin/hwctl.rs
+++ b/client/bin/hwctl.rs
@@ -21,8 +21,8 @@ use anyhow::{Context, Result};
 use clap::Parser;
 use std::process::ExitCode;
 
+use hwlib::check_certification_status;
 use hwlib::models::request_validators::{CertificationStatusRequest, Paths};
-use hwlib::send_certification_status_request;
 
 /// CLI tool to check hardware certification status.
 ///
@@ -58,10 +58,16 @@ fn main() -> ExitCode {
 }
 
 fn run(server_url: String) -> Result<()> {
-    let cert_status_request =
+    let current_hardware =
         CertificationStatusRequest::new(Paths::default()).context("cannot collect system data")?;
-    let response = send_certification_status_request(server_url, &cert_status_request)
-        .context("cannot send certification status request")?;
+
+    let response = check_certification_status(
+        server_url,
+        hwlib::CheckCertificationMode::Normal,
+        &current_hardware,
+    )
+    .context("Failed")?;
+
     let response_json =
         serde_json::to_string_pretty(&response).context("cannot serialize response as JSON")?;
     println!("{response_json}");
--- a/client/bin/hwctl.rs
+++ b/client/bin/hwctl.rs
@@ -65,6 +65,7 @@ fn run(server_url: String) -> Result<()> {
         server_url,
         hwlib::CheckCertificationMode::Normal,
         &current_hardware,
+        None,
     )
     .context("Failed")?;
 
--- a/client/bin/hwctl.rs
+++ b/client/bin/hwctl.rs
@@ -63,7 +63,7 @@ fn run(server_url: String) -> Result<()> {
 
     let response = check_certification_status(
         server_url,
-        hwlib::CheckCertificationMode::Forced,
+        hwlib::CheckCertificationSource::Server,
         &current_hardware,
         None,
     );
--- a/client/bin/hwctl.rs
+++ b/client/bin/hwctl.rs
@@ -66,8 +66,7 @@ fn run(server_url: String) -> Result<()> {
         hwlib::CheckCertificationMode::Forced,
         &current_hardware,
         None,
-    )
-    .context("Failed")?;
+    );
 
     let response_json =
         serde_json::to_string_pretty(&response).context("cannot serialize response as JSON")?;
--- a/client/bin/hwctl.rs
+++ b/client/bin/hwctl.rs
@@ -73,7 +73,7 @@ fn run(server_url: String) -> Result<()> {
         serde_json::to_string_pretty(&response).context("cannot serialize response as JSON")?;
     println!("{response_json}");
 
-    let (staled, stale_reason) = response.is_staled();
+    let (staled, stale_reason) = response.stale_status();
 
     if staled {
         return Err(anyhow::anyhow!(
--- a/client/src/cache.rs
+++ b/client/src/cache.rs
@@ -64,7 +64,7 @@ pub struct HWCache {
 #[derive(Serialize, Deserialize, Debug, Default)]
 struct HWCacheData {
     certification_status: CertificationStatus,
-    certification_status_url: Option<String>,
+    certification_certified_url: Option<String>,
     stale: StaleStatus,
     stale_reason: Option<String>,
     last_attempt_at: Option<String>,
@@ -73,6 +73,7 @@ struct HWCacheData {
     hardware_data: Option<CertificationStatusRequest>,
     remote_access_enabled: bool,
     server: Option<String>,
+    available_releases: Vec<models::software::OS>,
 }
 
 impl HWCache {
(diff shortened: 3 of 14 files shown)


sergio-costas:
@pedro-avalos This is a proposal for adding the cache fields. What do you think?

client/bin/hwctl.rs
pedro-avalos, line 1: I personally think this is too much logic to add to the `bin/hwctl.rs` side. The logic should likely reside in `hwlib` (i.e., `src/`).

client/src/models/response_validators.rs
pedro-avalos, line 1: Hm, this is not the right place to add these new fields. The `models/` code models and validates what the server expects and responds with. That's why you need all the `Option<...>` (some of these will not actually be optional, as per your spec).

I think we should either:

1. Create a `cache/` module where we can have `helpers.rs`, `modules.rs`, etc.
2. Since the cache fields are all shared (same no matter the server response), we could have a single `cache.rs` if the implementation will be relatively small

I think I lean towards 2 myself...

Then we'll update the following files:

1. `constants.rs`: A new `HWLIB_CACHE_DIR_PATH` with a default value. For now (since we have the deb) maybe it needs to be an `env!` like `OS_RELEASE_FILE_PATH`?
2. `cache.rs`: `CertificationStatusCache` structure (derive `Serialize` and `Deserialize`)
3. `cache.rs`: `Paths` structure (derive `Clone`) with two fields: `cache_filepath` and `request_filepath`
4. `cache.rs`: Implementation of `Paths` with function `default()` (see `src/models/request_validators.rs` for an example)
5. `cache.rs`: Implementation (some of these maybe could be functions outside of the impl) of `CertificationStatusCache` with functions: `new(paths)` to just populate based on cache; and `validate(paths)` to validate/stale cache
6. `lib.rs`: Some `CheckMode` enum with the modes specified in your spec
7. `lib.rs`: I think we need a `ClientResponse` or similar structure to define a specific syntax for the client response, which would have `response` and `cache` as fields. I'd like it, however, to be flat like in your spec; we could add a I think we could get away with `#[serde(flatten)]`, but I'm just thinking out loud at this point
8. `lib.rs`: A new function `check_certification_status(mode)` that uses the existing `send_certification_status_request` and new cache functionality
9. `bin/hwctl.rs`: Replace function being called with `check_certification_status`

And lastly: 10: Profit

Sorry, if this is a very long comment and feedback. I really do appreciate the work you've put into this, and I love the spec and way forward with Hardware API!

pedro-avalos (changes requested):
Thank you once again for the contribution and great ideas you've brought to Hardware API. I really appreciate the thought you've put behind your spec and willingness to contribute. I've laid out some feedback to make these changes in what I think could be a more structured form.

client/bin/hwctl.rs
sergio-costas: You are right. I'll give it a spin (hope that is the right expression...)

client/src/models/response_validators.rs
sergio-costas: No comment is very long, don't worry. Thanks for taking your time to answer it.

You are right: mixing the output with the cache data and the server response is wrong, and that having different structures for each of the three is the right way. Also, I agree in that solution 2 seems to be the best at this moment. In case that the cache gets too complex, we can do a refactor.

About 1): I think that we shouldn't merge any of these changes before we already have fully migrated to the snap. I already began sending a patch to add the snap to the ISO seeds, and prepared a plan/route. I'll send the links.

Also, about the ClientResponse, I think that it should contain only the needed info, and not the whole info sent by the server, to avoid exposing sensitive data. Basically what the spec already contains.

sergio-costas:
@pedro-avalos Completed a proposal for the cache. Can you check it when you have some spare time just to ensure that is what you want, please?

client/src/constants.rs
```diff
@@ -27,3 +27,7 @@ pub const PROC_VERSION_FILE_PATH: &str = "/proc/version";
 pub const OS_RELEASE_FILE_PATH: &str = env!("OS_RELEASE_FILE_PATH");
 
 pub const LSMOD: &str = "/usr/bin/lsmod";
+
+pub const CACHE_EXPIRATION_IF_CERTIFIED: u64 = 60 * 60 * 24 * 30; // 30  days
+pub const CACHE_EXPIRATION_IF_NOT_CERTIFIED: u64 = 60 * 60 * 24; // 1 day
+pub const CACHE_PATH: &str = "hw_cache.json";
```
pedro-avalos: This might be better named `CACHE_FILE_NAME`

client/Cargo.toml
```diff
@@ -34,6 +35,7 @@ pyo3 = { version = "0.28.0", features = ["extension-module"], optional = true }
 serde = { version = "~1.0.0", features = ["derive"] }
 serde_json = "~1.0.0"
 smbios-lib = "0.9.0"
+test-temp-dir = "0.7.0"
```
pedro-avalos: This probably should be in the `[dev-dependencies]` section

client/src/cache.rs
```diff
@@ -0,0 +1,483 @@
...
+use std::path::Path;
+use std::{env, fs::File};
+
+use chrono::DateTime;
+use serde::{Deserialize, Serialize};
+
+#[cfg(test)]
+use crate::models::software::OS;
+#[cfg(test)]
+use test_temp_dir::{test_temp_dir, TestTempDir};
```
pedro-avalos: Can you move these to the `#[cfg(test)]` blocks where they are used instead of having them at the top of this module?

client/src/cache.rs
```diff
@@ -0,0 +1,483 @@
...
+    fn default() -> Self {
+        StaleStatus::Valid
+    }
+}
+
+/// A cache for the hardware data and certification status.
+pub struct HWCache {
+    data: HWCacheData,
+    current_hardware_data: Option<CertificationStatusRequest>,
+    cache_path: String,
```
pedro-avalos: This should probably be a `PathBuf`

client/src/cache.rs
```diff
@@ -0,0 +1,483 @@
...
+    last_attempt_at: Option<String>,
+    checked_at: Option<String>,
+    expires_at: Option<String>,
+    hardware_data: Option<CertificationStatusRequest>,
+    remote_access_enabled: bool,
+    server: Option<String>,
+}
+
+impl HWCache {
+    fn get_cache_path(cache_folder: Option<String>) -> String {
```
pedro-avalos: I think this should take in a `Option<&Path>` and return a `PathBuf`.

For consistency, also, let's use explicit `return` statements.

client/src/cache.rs
```diff
@@ -0,0 +1,483 @@
...
+        }
+        let cache_path = Path::join(Path::new(&snap_data), crate::constants::CACHE_PATH);
+        cache_path.to_str().unwrap().to_string()
+    }
+
+    fn get_now(&self) -> DateTime<chrono::Utc> {
+        chrono::Utc::now()
+    }
+
+    pub fn new(cache_folder: Option<String>) -> Self {
```
pedro-avalos: Same as before, this should take in a `Option<&Path>` in my opinion

client/src/cache.rs
```diff
@@ -0,0 +1,483 @@
...
+        (
+            self.data.certification_status.clone(),
+            self.data.stale.clone(),
+            self.data.stale_reason.clone(),
+        )
+    }
+
+    /// Compares the given hardware data with the cached hardware data.
+    /// Returns true if they are the same, false otherwise.
+    pub fn compare_hardware_data(&mut self, hardware_data: &CertificationStatusRequest) -> bool {
```
pedro-avalos: Does `self` really need to be `&mut`? I think `&self` is enough.

client/src/lib.rs
```diff
@@ -36,6 +37,141 @@ use models::{
...
+    valid_cache: bool,
+    hardware_mismatch: bool,
+    stale: bool,
+    stale_reason: Option<String>,
+    source: CertificationSource,
+    remote_access_enabled: bool,
+}
+
+fn create_answer(
+    cache: &mut HWCache,
```
pedro-avalos: I think once `compare_hardware_data` changes from `&mut self` to `&self`, this can also be changed to `&HWCache`

client/src/lib.rs
```diff
@@ -36,6 +37,141 @@ use models::{
...
+    source: CertificationSource,
+    hardware_info: &CertificationStatusRequest,
+) -> PublicCertificationStatus {
+    let (certification_status, stale_status, stale_reason) = cache.get_status();
+
+    let hardware_mismatch = !cache.compare_hardware_data(hardware_info);
+
+    PublicCertificationStatus {
```
pedro-avalos: Nitpick: I don't think we need a blank line between these statements

client/src/lib.rs
```diff
@@ -36,6 +37,141 @@ use models::{
...
+        CertificationStatusResponse::CertifiedImageExists { .. } => {
+            certification_status = CertificationStatus::Certified;
+        }
+        CertificationStatusResponse::RelatedCertifiedSystemExists { .. } => {
+            certification_status = CertificationStatus::RelatedCertifiedSystemExists;
+        }
+        _ => {
+            certification_status = CertificationStatus::NotSeen;
+        }
+    }
```
pedro-avalos, line 208: I think this logic could be an implementation of `from` for `CertificationStatus`. I think something like the following:

```rust
impl From<CertificationStatusResponse> for CertificationStatus {
    fn from(request: CertificationStatusResponse) -> Self {
        match request {
            CertificationStatusResponse::Certified { .. } => CertificationStatus::Certified,
            CertificationStatusResponse::CertifiedImageExists { .. } => {
                CertificationStatus::CertifiedImageExists
            }
            CertificationStatusResponse::RelatedCertifiedSystemExists { .. } => {
                CertificationStatus::RelatedCertifiedSystemExists
            }
            _ => CertificationStatus::NotSeen,
        }
    }
}
```

Then this line can just be `certification_status = CertificationStatus::from(response);`

client/src/lib.rs
```diff
@@ -36,6 +37,141 @@ use models::{
...
+) -> Result<PublicCertificationStatus> {
+    let mut cache = HWCache::new(None);
+
+    if mode == CheckCertificationMode::Cached {
+        return Ok(create_answer(
+            &mut cache,
+            CertificationSource::Cache,
+            hardware_info,
+        ));
+    }
```
pedro-avalos: Since we repeat the `create_answer` with the same variables, I think we could define the following right after `let mut cache = ...`:

```rust
let cache_answer = |cache: &HWCache| {
    Ok(create_answer(
            cache,
            CertificationSource::Cache,
            hardware_info,
    ))
};
```

Then when the answer is a cached answer, we just `return cache_answer(&cache);`

pedro-avalos (changes requested):
This looks much better, thanks! I have left some comments, but I think the overall structure and logic looks good to me.

client/src/constants.rs
sergio-costas: Done.

client/Cargo.toml
sergio-costas: Done

client/src/cache.rs
sergio-costas: Good point! Done.

client/src/cache.rs
sergio-costas: Good catch! Done.

client/src/lib.rs
sergio-costas: True.

client/src/lib.rs
sergio-costas: But I'll need to add some extra fields, like the URL that points to the certified device. I was going to do it today.

client/src/lib.rs
sergio-costas: Nice!!!! Thanks!

client/src/cache.rs
sergio-costas: All done. And you are right: the code is more elegant now. Thanks!

client/src/cache.rs
sergio-costas: Done.

client/src/lib.rs
sergio-costas: Added the URL.

client/src/py_bindings.rs
pedro-avalos, line 33: I think we should keep this function essentially as it was (i.e., a Python equivalent of `send_certification_status_request`) and add a new Python equivalent of `check_certification_status`.

pedro-avalos (changes requested):
One last comment to avoid breaking the Python bindings, but otherwise looks great! Thanks!

client/src/py_bindings.rs
sergio-costas: Something like this...?

sergio-costas:
@pedro-avalos Ready for review.

pedro-avalos (approved):
LGTM. Thank you so much for your contribution!

PR #581, opened 2026-06-26
Add formatting check instructions

sergio-costas:
## Description

Adds instructions about how to configure the pre-commit hook in git to check the code formatting.


pedro-avalos:
I like git hooks, but I think recommending some git hook snippets in `CONTRIBUTING.md` will not scale very well or be very maintainable. What do you think about #587?

sergio-costas:
Mmm... yes, that looks good!

PR #582, opened 2026-06-26
Move snapcraft.yaml file to the root folder

sergio-costas:
This allows to build the snap directly from the root, and thus not needing to specify a directory. This is useful for launchpad recipes and for building the snap directly from the snap store.


pedro-avalos:
Hm, I'm not sure I like this change, as it breaks the monorepo structure of the project.

sergio-costas:
I'll close it because it's not required anymore.

PR #596, opened 2026-07-02
Daemonize hwctl

sergio-costas:
This PR moves the code to a socket-activated daemon, and makes the client connect to the socket to ask for the data using VarLink for the wire protocol. This removes the need for root access.

It also adds several tests, including an end-to-end for the snapped service and command-line client.

## Testing

Built the snap, installed it, and run `hwctl` as a normal user. It returned the expected data.

## Checklist

- [x] I have added necessary tests.
- [x] I have added or updated any relevant documentation (if needed).


sergio-costas:
@pedro-avalos Ok, this is ready for the first review.

sergio-costas:
@pedro-avalos I've been thinking... Currently, the Varlink API is very simple, and it just returns a string with an embedded JSON with the same data that we previously returned through the hwctl command line tool. But I'm thinking if, maybe, we should take advantage of the Varlink protocol itself and return the data directly in the call. This is, instead of:

```
type CertificationSource (auto, cache, server)

method GetCertificationStatus(
    source: CertificationSource,
    server_url: ?string
) -> (status: string)
```

use:

```
type CertificationSource (auto, cache, server)
type CertificationStatus (Unknown, Certificated, NotSeen, CertifiedImageExists, RelatedCertifiedSystemExists)
type Kernel (
    name: ?string,
    version: string,
    signature: ?string,
    loaded_modules: []string
)
type OS (
    codename: string,
    distributor: string,
    version: string,
    kernel: Kernel
)
type State (
    status: CertificationStatus,
    certified_url: ?string,
    available_releases: ?[]OS,
    valid_cache: boolean,
    hardware_mismatch: boolean,
    stale: boolean,
    stale_reason: ?string,
    source: CertificationSource,
    remote_access_enabled: boolean,
    server_url: string
)

method GetCertificationStatus(
    source: CertificationSource,
    server_url: ?string
) -> (
    error: ?string,
    state: ?State
)
```

But I don't know if it would be making it too complex. Maybe a compromise solution, like:

```
method GetCertificationStatus(
    source: CertificationSource,
    server_url: ?string
) -> (
    error: ?string,
    status: CertificationStatus,
    certified_url: ?string,
    available_releases: ?string,
    valid_cache: boolean,
    hardware_mismatch: boolean,
    stale: boolean,
    stale_reason: ?string,
    source: CertificationSource,
    remote_access_enabled: boolean,
    server_url: string
)
```

(thus, just setting the available releases as a string with a JSON inside...).

I'm mainly afraid of having trouble if we update the format in the server side in the future...

sergio-costas:
Nah, I decided to go with it.

.github/workflows/test_client.yaml
pedro-avalos, line 1: I'm okay with this, I think we might want to look at adding some more formal tests in the future, though, maybe via spread?

client/bin/com.ubuntu.hwctl.varlink
pedro-avalos, line 1: I'm not very familiar with varlink, but this looks good; I do prefer the explicit agreement/signature

README.md
```diff
@@ -11,7 +11,7 @@ check the certification status of hardware configurations.
 A full deployment of Hardware API consists of the following components:
 
 - [`hwapi`](./server/): The Hardware API server.
-- [`hwctl`](./client/): The Hardware API client (library and CLI tool).
+- [`hwctl`](./client/): The Hardware API client (service and CLI tool).
```
pedro-avalos: Technically `hwlib` (library) is still in the `client/` directory


```suggestion
- [`hwctl`](./client/): The Hardware API client (library, CLI, and service).
```

client/snapcraft.yaml
```diff
@@ -28,14 +28,22 @@ platforms:
   amd64:
 
 apps:
-  hwctl:
-    command: bin/hwctl
+  hwctl-daemon:
```
pedro-avalos, line 31: I wonder if we should name this `hwctld`, or if the `*-daemon` is more standard. I'm okay with either, though.

Also, I don't expect we'll need a snap alias, right? This is a daemon we expect to be handled by the OS/snap and not directly by the user. Is this correct?

client/snapcraft.yaml
```diff
@@ -28,14 +28,22 @@ platforms:
...
     plugs:
       - hardware-observe
       - kernel-module-observe
       - network
+      - network-bind
       - network-setup-observe
       - system-observe
+    daemon: simple
+    sockets:
+      hwctlsocket:
```
pedro-avalos: This might just be me, but I think this should be `hwctl-socket` with a dash?

client/snapcraft.yaml
```diff
@@ -28,14 +28,22 @@ platforms:
...
+        listen-stream: $SNAP_COMMON/hwctl.varlink
+
+  hwctl:
+    command: bin/hwctl
```
pedro-avalos, line 46: Interesting, the CLI doesn't need *any* plugs now? I'm not very familiar with sockets and services in snapd, but we only need to define something in the daemon-side?

If so, cool

client/README.md
```diff
@@ -9,28 +9,61 @@
...
 
-It consists of both the `hwlib` library and `hwctl` CLI tool.
+It consists of both the `hwctl-daemon` service and `hwctl` CLI tool.
 
-- `hwlib`: Rust library for communicating with
-  [the API server](../server/).
-- `hwctl`: CLI tool leveraging the `hwlib` library.
+- `hwctl-daemon`: Rust service for communicating with
+  [the API server](../server/) and cache results.
+- `hwctl`: CLI tool to retrieve the certification status and other data.
```
pedro-avalos: Again, `hwlib` is still a thing, so I'd say let's leave it in the readme, along with the CLI and daemon

client/README.md
```diff
@@ -47,6 +80,42 @@ On Questing Quokka (25.10), you can also install it using `apt`:
 sudo apt-get install hwctl
 ```
 
+## Service protocol
+
+The hwctl service is socket-activated, and it uses [Varlink](https://varlink.org/)
```
pedro-avalos: Could you put the `[varlink]` link in the bottom of the markdown along with the rest of the links?


```suggestion
The hwctl service is socket-activated, and it uses [Varlink]
```

client/README.md
```diff
@@ -9,28 +9,61 @@
...
-  [the API server](../server/).
-- `hwctl`: CLI tool leveraging the `hwlib` library.
+- `hwctl-daemon`: Rust service for communicating with
+  [the API server](../server/) and cache results.
+- `hwctl`: CLI tool to retrieve the certification status and other data.
 
 ## Basic Usage
 
-To check the machine certification status, simply run `hwctl` with
-root[^sudo] privileges:
```
pedro-avalos, line 21: If the `[^sudo]` footnote is not necessary anymore, let's delete it at the bottom of the readme

pedro-avalos (changes requested):
Looks great! Thank you for adding this and all the testing; I have left some comments regarding documentation and naming

README.md
sergio-costas: Ops... good point! Changing it.

client/snapcraft.yaml
sergio-costas: Good point. But in my opinion, `hwctld` is too much near to `hwctl`, so my idea was to ensure that there is no possible confusion. Also... maybe `hwctlXXXXX` isn't a good idea because of the `ctl` part...

Maybe `hardwared`? `certification-service`? Or `hwctl-service`?

client/snapcraft.yaml
sergio-costas: Yes! Done.

client/snapcraft.yaml
sergio-costas: It works without any plug because it is running in the same snap than the daemon/service. To allow other snapped programs to access this, it would be required either using the [contents](https://snapcraft.io/docs/reference/interfaces/content-interface/#interfaces-content-interface) interface, or to create a specific interface for this, to allow that program to read and write the Varlink socket (which is placed at `/var/snap/hwctl/common/hwctl.varlink`). That's all that is required to get the certification status. I would go for the second one, because it could be unprivileged.

sergio-costas: For non-snapped programs, like gnome-control-center, no specific privileges are required. In fact, I [completed a patch](https://salsa.debian.org/gnome-team/gnome-control-center/-/merge_requests/88) to show the certification status in gnome-control-center, which uses the code autogeneration offered by Varlink to get the data and show it, and doesn't require root access or any kind of permissions anymore.

client/README.md
sergio-costas: Mmm... but the point is that, now, `hwlib` is not really needed anymore... You just connect to the Varlink socket and get the data. And since it's a protocol with an IDL and reflection support, static languages use code generators (like `varlink-generator` for Rust, or `vali_compiler` for C), and dynamic languages (like python) just connect to the socket and get all the required data from there to build the interface in run time. So... which uses remain for `hwlib`?

client/README.md
sergio-costas: Ok.

client/README.md
sergio-costas: Good point! Done.

.github/workflows/test_client.yaml
sergio-costas: It sounds good. Agree.

client/snapcraft.yaml
pedro-avalos: I think you're right, and `hwctl-daemon` makes sense

client/README.md
pedro-avalos: It is still published in crates.io, but more generally, `hwlib` is the name of the client library that is the backend for `hwctl` (and now `hwctl-daemon`).

sergio-costas: Good point. Fixed. What about now?

client/src/lib.rs
```diff
@@ -78,6 +79,19 @@ impl PublicCertificationStatus {
...
+    pub fn extra_data(&self) -> (bool, CertificationSource, bool, String, bool) {
+        return (
+            self.valid_cache,
+            match self.source {
+                CertificationSource::Cache => CertificationSource::Cache,
+                CertificationSource::Server => CertificationSource::Server,
+            },
+            self.remote_access_enabled,
+            self.server_url.clone(),
+            self.remote_access_enabled,
```
pedro-avalos: Is this duplicated `self.remote_access_enabled` a typo? I thin kit's supposed to be `self.hardware_mismatch`?

client/bin/hwctl-daemon.rs
```diff
@@ -0,0 +1,278 @@
...
+        };
+        return call.reply(response_varlink);
+    }
+
+    fn set_remote_access(
+        &self,
+        call: &mut dyn com_ubuntu_hwctl::Call_SetRemoteAccess,
+        r#enabled: bool,
+    ) -> varlink::Result<()> {
+        let mut cache = hwlib::cache::HWCache::new(None);
```
pedro-avalos, line 164: Comment from Copilot review (I think this one is a false flag, but not sure):

> `client/bin/hwctl-daemon.rs` uses `HWCache::new(None)`, which resolves via `SNAP_DATA` (falling
     back to `"."`/CWD if unset). The snap declares its socket under `$SNAP_COMMON` but the cache
     uses `SNAP_DATA`. Needs verification that `SNAP_DATA` is always populated for the daemon and
     that the write lands in the same file `check_certification_status` later reads.

client/bin/hwctl-daemon.rs
```diff
@@ -0,0 +1,278 @@
...
+                hwlib::cache::CertificationStatus::Unknown => CertificationStatus::Unknown,
+                hwlib::cache::CertificationStatus::CertifiedImageExists => {
+                    CertificationStatus::CertifiedImageExists
+                }
+                hwlib::cache::CertificationStatus::RelatedCertifiedSystemExists => {
+                    CertificationStatus::RelatedCertifiedSystemExists
+                }
+            },
+            certified_url: url,
+            available_releases: None,
```
pedro-avalos: Is this intended to be `None`?

client/src/helpers.rs
```diff
@@ -16,14 +16,62 @@
...
+fn check_path_exists(socket_path: &str) -> bool {
+    let socket_path = std::path::Path::new(&socket_path);
+    return socket_path.exists();
+}
+
+pub fn get_socket_path(binary_type: BinaryType) -> Result<(String, String), anyhow::Error> {
+    if std::env::var("HWCTL_SOCKET_PATH").is_ok() {
+        let path = std::env::var("HWCTL_SOCKET_PATH").unwrap();
+        return Ok((path, std::env::var("HWCTL_SOCKET_PATH").unwrap()));
+    }
```
pedro-avalos: I think this could be refactored to read `HWCTL_SOCKET_PATH` once:

```rust
if let Ok(path) = std::env::var("HWCTL_SOCKET_PATH") {
    return Ok((path.clone(), path));
}
```

client/bin/hwctl-daemon.rs
```diff
@@ -0,0 +1,278 @@
...
+    );
+
+    let _ = varlink::listen(
+        service,
+        &socket_file,
+        &varlink::ListenConfig {
+            idle_timeout: timeout,
+            ..Default::default()
+        },
+    );
```
pedro-avalos: I think this should be returning the varlink listen result or some form of error/success.

client/bin/hwctl-daemon.rs
```diff
@@ -0,0 +1,278 @@
...
+                "ERROR: failed to create socket directory {}: {:?}",
+                full_path.display(),
+                e
+            );
+            return ExitCode::FAILURE;
+        }
+    }
+
+    create_server(socket_file, 60);
+    return ExitCode::SUCCESS;
```
pedro-avalos: And here the return should be `FAILURE` or `SUCCESS` depending on the `create_server` result

client/src/lib.rs
pedro-avalos: This should probably have `Clone` and `Copy` to simplify some code down in this file

client/src/lib.rs
```diff
@@ -78,6 +79,19 @@ impl PublicCertificationStatus {
...
         return (self.stale, self.stale_reason.clone());
     }
+
+    pub fn extra_data(&self) -> (bool, CertificationSource, bool, String, bool) {
+        return (
+            self.valid_cache,
+            match self.source {
+                CertificationSource::Cache => CertificationSource::Cache,
+                CertificationSource::Server => CertificationSource::Server,
+            },
```
pedro-avalos: This can become `self.source` with the `Clone` and `Copy`

client/bin/hwctl-daemon.rs
```diff
@@ -0,0 +1,278 @@
...
+        &self,
+        call: &mut dyn com_ubuntu_hwctl::Call_SetRemoteAccess,
+        r#enabled: bool,
+    ) -> varlink::Result<()> {
+        let mut cache = hwlib::cache::HWCache::new(None);
+        if r#enabled {
+            cache.set_remote_access_enabled(true);
+        } else {
+            cache.set_remote_access_enabled(false);
+        }
```
pedro-avalos: This is fine, but could be a one-liner, no?


```suggestion
        cache.set_remote_access_enabled(r#enabled);
```

pedro-avalos (changes requested):
Thank you! I have left some last set of comments assisted by a quick run-down from copilot locally.

client/src/lib.rs
sergio-costas: It's a typo, yes. Fixed.

client/bin/hwctl-daemon.rs
sergio-costas: Yes, they must be different to ensure that the cache is updated if the snap is updated or downgraded.

client/bin/hwctl-daemon.rs
sergio-costas: Oh... I didn't implement that... give me a moment.

client/src/helpers.rs
sergio-costas: Elegant...

client/bin/hwctl-daemon.rs
sergio-costas: Done.

client/src/lib.rs
sergio-costas: Do you mean to use the Varlink definitions for the cache too?

client/src/lib.rs
sergio-costas: Oh... good point.

client/src/lib.rs
sergio-costas: Got it, and done.

client/src/lib.rs
sergio-costas: Done.

client/bin/hwctl-daemon.rs
sergio-costas: Yes. Done.

client/bin/hwctl-daemon.rs
sergio-costas: Done.

client/bin/hwctl-daemon.rs
sergio-costas: Done.

sergio-costas:
@pedro-avalos Ok, all done. But @aleasto commented a problem: allowing a snap without network access, but with access to the hardware service, to specify the URL for the server, is a breach, so I decided to add a snap parameter (`allow-custom-url`) that, when set to true (or 1), allows any connected client to pass a custom URL, and if not set (or set to false), fails if a custom URL is passed.

The problem is how to access that snap parameter: currently I'm running the `snapctl get allow-custom-url` command every time a petition is done. Do you know if there is a Rust API to read that?

EDIT: no, it doesn't exist. The `snapctl` command is the only API allowed for that.

client/README.md
```diff
@@ -47,6 +82,41 @@ On Questing Quokka (25.10), you can also install it using `apt`:
...
+
+The hwctl service is socket-activated, and it uses [Varlink] as the underlying protocol.
+The protocol is defined in the `bin/com.ubuntu.hwctl.varlink` file. It has two methods:
+
+* GetCertificationStatus(source: CertificationSource, server_url: ?string) -> (state: State)
+  It receives a `source` parameter, which can be any of `server`, `cache` or `auto`,
+  and an optional (thus, it can be `null`) string with the certification server URL.
+  It returns a State object with these fields:
+
+  * status: `NotSeen`, `Certificated`, `CertifiedImageExists`, `RelatedCertifiedSystemExists` or `Unknown`.
```
pedro-avalos: I think this is a typo for `Certified`, right?

```suggestion
  * status: `NotSeen`, `Certified`, `CertifiedImageExists`, `RelatedCertifiedSystemExists` or `Unknown`.
```

client/README.md
```diff
@@ -47,6 +82,41 @@ On Questing Quokka (25.10), you can also install it using `apt`:
...
+  * certified_url: either `null` if there is no applicable URL, or the URL describing either the current
+    hardware, or the URL describing the related certified system.
+  * available_releases: an array with the certified OS images available for this device, in the case of
+    `CertifiedImageExists`. If not, it will be `null`.
+  * valid_cache: a boolean that specifies whether the cache is valid or is old/not-existen/other problem.
```
pedro-avalos:
```suggestion
  * valid_cache: a boolean that specifies whether the cache is valid or is old/non-existent/other problem.
```

client/README.md
```diff
@@ -47,6 +82,41 @@ On Questing Quokka (25.10), you can also install it using `apt`:
...
+  * hardware_mismatch: a boolean that specifies if the current hardware has changed respect to the one
```
pedro-avalos:
```suggestion
  * hardware_mismatch: a boolean that specifies if the current hardware has changed in respect to the one
```

client/README.md
```diff
@@ -47,6 +82,41 @@ On Questing Quokka (25.10), you can also install it using `apt`:
...
+  * remote_access_enabled: a boolean specifying if the service should refresh the cached data when
+    using the `auto` source, or not.
+  * server_url: a string with the server URL from which all the shown data was obtained in origin.
+
+* SetRemoteAccess(enabled: bool) -> ()
+  It receives a single boolean that changes the `remote_access_enabled` setting inside the service.
+
+For snapped services, the socket path is `/var/snap/hwctl/common/hwctl.varlink`, while for
+unconfined services, it is `/run/hwctl/hwctl.varlink`. Clients must always check the former
+first, and only check for the later if the former doesn't exists.
```
pedro-avalos:
```suggestion
first, and only check for the latter if the former doesn't exists.
```

client/test_utils/certification_emulator.py
```diff
@@ -0,0 +1,219 @@
...
+      "kernel": {
+        "name": null,
+        "version": "6.8.0-1009-oem",
+        "signature": null,
+        "loaded_modules": []
+      }
+    }
+  ],
+  "pci_peripherals": [],
+  "usb_peripherals": [],
```
pedro-avalos, line 198: Technically trailing commas are bad JSON

```suggestion
  "usb_peripherals": []
```

pedro-avalos (approved):
Just typos, but the code looks good! Thank you!!!

client/README.md
sergio-costas: Agh! Yes!

client/test_utils/certification_emulator.py
sergio-costas: That's the point, it's a malformed JSON. I'll add a comment.

sergio-costas:
@pedro-avalos Ok, merged your comments, but also did a change to manage the snap settings TheWayItShouldBe(tm). Now the configuration is stored in a separate file, and is modified by the `configure` hook (located at `snap/hooks`). This way it's not needed to run `snapctl` every time the daemon connects to the server.

The changes are in this commit: https://github.com/canonical/hardware-api/pull/596/changes/2bb6d5adce16813c3b955a35278187531495d577 This should be the last change.

pedro-avalos (approved):

PR #608, opened 2026-07-24
fix(client): resolve lint warnings

pedro-avalos:
## Description

- This PR fixes the linter warnings we are seeing in CI and locally

## Testing

- `just client::lint`

## Checklist

- [X] I have added necessary tests.
- [X] I have added or updated any relevant documentation (if needed).

## Additional Notes

[contribution guidelines]: https://github.com/canonical/hardware-api/blob/main/CONTRIBUTING.md
[cla]: https://ubuntu.com/legal/contributors

### DIFF
--- a/client/.gitignore
+++ b/client/.gitignore
@@ -1,3 +1,4 @@
+bin/com_ubuntu_hwctl.rs
 target
 cargo_home
 .cargo-checksum.json
--- a/client/bin/hwctl-daemon.rs
+++ b/client/bin/hwctl-daemon.rs
@@ -96,6 +96,7 @@ impl VarlinkInterface for ComUbuntuHwctl {
         if current_hardware.is_err() {
             return call.reply_cannot_collect_system_data();
         }
+        #[allow(clippy::unnecessary_literal_unwrap)]
         let current_hardware = current_hardware.unwrap();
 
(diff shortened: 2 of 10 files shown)


client/.gitignore
```diff
@@ -1,3 +1,4 @@
+bin/com_ubuntu_hwctl.rs
```
pedro-avalos, line 1: @sergio-costas this seems to be auto-generated and wasn't included in your PR, is it safe to gitignore if it's not meant to be checked in?

sergio-costas: Yes, it is safe.

sergio-costas:
So... we are moving the code from "always use return" to "use return only where required"...?

pedro-avalos:
> So... we are moving the code from "always use return" to "use return only where required"...?

@sergio-costas seems like the lint rules have changed :man_shrugging: 

We could probably define them in a config file, I suppose. I am personally used to Python's always use return, so that's where I was coming from, but maybe Rust has its own rules

sergio-costas (approved):
Ok, it looks fine, although, personally, would prefer to keep some of those returns.

sergio-costas:
Don't forget to squash the commits, please.

